Issue #7, opened 2019-03-22
Views


EdwardMoyse:
Add some preset views, and animate between them.

9inpachi:
Hi Edward,

Does it have to be like this?
- There is a list of preset views in the controls section
- When user clicks on one of the presets, the camera, in an animating fashion shifts to that view
- A view can be inside some part of the detector, or the origin of particles or other related parts of an event display

ric-bianchi:
Hi @9inpachi , yes, that's an option. Default views will be used by users to create "standard" event displays more easily.

9inpachi:
Hi,
I have tried implementing views a bit.

Here's a screenshot of what I have done so far:
![presetViews](https://user-images.githubusercontent.com/36920441/55718826-22bb6e80-5a16-11e9-992d-b472855626d1.png)

You can test and see the changes at: https://github.com/9inpachi/phoenix
The latest commit `Preset Views Test` shows the changes done. The shifting of views is not animated yet, it is just a test to know if I am going in the right direction.

Please let me know if this is right.

EdwardMoyse:
Very nice! I think we might ultimately want to do this with icons rather than using the pulldown GUI, but having both is good too. 
Feel free to make a pull request!

9inpachi:
I will add the icons and make a pull request then. :)

9inpachi:
Hi,
I have added animations for preset views using tween js. Before I make a pull request, I wanted to know if its alright to include a new library like tween js?

Here's the output:
![presetViews](https://user-images.githubusercontent.com/36920441/56083150-9dcbbd00-5e3a-11e9-8484-97603f4f4198.gif)

Thanks. :-)

EdwardMoyse:
That's absolutely fine!

PR #9, opened 2019-04-11
Preset Views

9inpachi:
Addresses: #7 

Hi,
I have tried adding icon support to preset views. The icons appear at the middle top.

preset-icons

I am not sure if fontawesome icons are doing a good enough job. Should we use custom made images instead?
However, what do you think of it so far?


EdwardMoyse:
I think that's great! We can always update the icons later, but it's a very nice start.

PR #14, opened 2019-05-15
Fixing broken link

emiliocortina:


EdwardMoyse:
Thanks! I actually fixed this myself yesterday.

This is my lack of understanding of github I guess (I'm used to gitlab) but I don't really understand why this PR isn't showing that there is no difference with what is in master?

Issue #3, opened 2019-03-22
Refactor code


EdwardMoyse:
We should split up jsdisplay.js into smaller fragments, such as tracks, jets etc.

PR #17, opened 2019-06-08
[Refactoring] Angular port 

emiliocortina:
Hi! After a bunch of issues with Github branches it finally lets me open a pull request.
This is the refactoring #3  into components I did using Angular, it is actually very easy once you figure out how it is structured as I didn't use many Angular specific features.
I will explain it to you the next time we talk more in detail but the overall structure is this:

- Inside src/app you will find the different folders containing the components (atlas, geometry, trackml, home), these folders have a .html, .css and .ts  files (the other .spec.ts is for unit testing). This is more or less equivalent as having embedded css and javascript (ts files are very similar to js files, you can even use JavaScript syntax in them) in the html documents (like in the previous phoenix_TrackML.html).

- Then the other main part is inside /src/app/services. These services encapsulate functionality. The main one is EventDisplayService, this is the one that components call, sort of an intermediate, that then delegates making calls to UIService (encapsulates the functionality related to the onscreen controls) and ThreeServices (encapsulates the functionality related to managing the library ThreeJS).

Finally, another thing that's changed is that we don't have to manually download and add the different .js files containing the libraries, everything is done importing node modules, and you can see how they are declared for example in the first lines of /src/app/services/three.service.ts.

Hope it is a bit clearer! :)


EdwardMoyse:
Wow this is fantastic Emilio! It looks like there are some conflicts ... but I will have a proper look today.

PR #30, opened 2019-07-25
Added the ability to switch between orthographic and perspective view.

ss4774:
For enhancement purposes I have added the ability to switch between orthographic and perspective view. I have also included some new controls in the GUI. The controls include a toggle to switch camera view mode and additional alignment options.

### DIFF
--- a/src/app/services/three.service.ts
+++ b/src/app/services/three.service.ts
@@ -12,7 +12,8 @@ import {
   MeshBasicMaterial, Object3D,
   PerspectiveCamera,
   Scene, Vector3,
-  WebGLRenderer
+  WebGLRenderer,
+  OrthographicCamera,
 } from 'three';
 import {Configuration} from './loaders/configuration.model';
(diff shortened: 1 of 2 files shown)


EdwardMoyse:
Very nice!

As I said in a separate conversation, I'm not sure we need the 'align' buttons, since we have alignment defined elsewhere. But let's see what others think.

There is one issue with  the clip plane being too close to the geometry now. See image below. But let's merge this for the moment and fix that later.

image

ss4774:
I was undecided about alignment as well, but I let them be since they work a little bit different.

I think this bug is due to the orthographic camera not moving back and forth when zoom is applied. I believe three.js in this way simulates zoom with orthographic camera, since zoom should not affect an orthographic projection. As a result camera stays stationary at fixed position and therefore its clipping planes.

I am currently trying to bypass this behavior while working on overlay (image below) and hopefully I will find a way around this.
Screenshot from 2019-07-29 20-16-05

PR #32, opened 2019-08-17
General improvements and fixes

ss4774:
- Improved export and import 
Here I tackle odd behavior while exporting and importing a scene. I introduce list of object to not export, since they are not well supported in current GLTF versions. The objects that are excluded are considered a defaults and are loaded at scene initialization.

- Fixed views animation to work on arbitrary transform 
Views animation did not seem to work in arbitrary position and rotation. Adding both animations fixed the problem. I also added animation to other align view options.

- Fixed muons color property
There was an error while tiring to change a muon color. Since muons are compound objects, changing the depth of material color assignment solved this issue. 

-  Formatting and fixing event data loading
Changed event data loading to a more consistent format. Please note that I also changed a sorting function of track positions (file: phoenix-loader.ts, lines: 125 - 130). I think there was an error in the sorting function, but it may be intentionally that way.
Maybe we should exclude a sort completely since sorting might corrupt a correct order of track positions. Trajectory of a particle may bend in such a way that it loops back to collision source and sorting would give incorrect order of track positions. This seems unlikely but the probability is still there.

- Fixing selecting items on reimported scene
Alternative solution to enable selection on object in a scene. Solved by adding an even data at scene load.

### DIFF
--- a/src/app/services/eventdisplay.service.ts
+++ b/src/app/services/eventdisplay.service.ts
@@ -77,13 +77,9 @@ export class EventdisplayService {
    */
   public loadEventsFromJSON(eventsData: any): string[] {
     this.eventsData = eventsData;
-    const eventKeys = [];
-    for (const ev of Object.keys(eventsData)) {
-      if (eventsData[ev] != null) {
-        eventKeys.push(ev);
-      }
-    }
+    const eventKeys = this.configuration.getEventDataLoader().buildEventsList(eventsData);
     this.loadEvent(eventKeys[0]);
+
     return eventKeys;
   }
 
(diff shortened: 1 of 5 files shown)


emiliocortina:
Everything looks great!
I cannot tell you about the sorting function in Phoenix Loader as the code for adding tracks, jets and muons was already done by @EdwardMoyse .

Another thing is in EventDisplayService:94, when you check that an event is an Array:

Captura de pantalla 2019-08-18 a las 11 16 53

In our current format (defined in Phoenix Loader) each event is always an object containing object types and inside them collections, however, if at some point another format representing events as arrays were to appear it should be defined in another loader, leaving the EventDisplayService to only manipulate the multiple events, and delegating on the loader to process the format of each particular event.

Please correct me if I misunderstood something.
Thank you!

ss4774:
I used that piece of code to read provided event data while I was trying to add compound objects (at that time I did not yet realize that you have already completed that issue).
Event data file was provided in compound objects issue declaration ([https://emoyse.web.cern.ch/emoyse/EventData.json](https://emoyse.web.cern.ch/emoyse/EventData.json)).

Event objects in that file seem to be wrapped in an array as a single element:
Screenshot from 2019-08-18 16-52-53_event
I was suspecting that the format was still changing so I left that code, which works on both formats.

Best regards!

PR #33, opened 2019-08-22
WIP Initial test to add the LHCb Geometry to Phoenix

bcouturi:
This is an attempt at loading the LHCb detector in glTF format into Phoenix.
I add problems loading the geometry with the existing loadGLTF method, so I worked around it for the moment (but this needs resolving before an eventual merge).


EdwardMoyse:
Hmm. Not quite sure why the CI failed!

PR #51, opened 2020-02-20
Use minimal classes for homepage header

9inpachi:
Hi,

Some of the classes being used for the homepage pricing-header can be removed without affecting its responsive behavior.


EdwardMoyse:
Perfect. Thanks a lot!

Issue #35, opened 2019-10-16
Make users aware of hardware acceleration


9inpachi:
Hi,
As hardware acceleration can significantly improve the performance for each event display. It would help the users to know if they have hardware acceleration enabled or disabled.
Unfortunately, there isn't a way to automatically request hardware acceleration from somewhere in the code but the performance and certain factors can be measured to know if hardware acceleration is being used.

Something like https://gist.github.com/cvan/042b2448fcecefafbb6a91469484cdf8 can also be used to detect the GPU being used by the renderer.
With hardware acceleration disabled, Chrome makes use of `Chrome SwiftShader`.
Since the behavior is different for each browser, I am not sure if this is a reliable way.

A manual way would be to let the users know in the form of a one-time popup that hardware acceleration can be used if they experience performance issues, redirecting the users to links for enabling it in each browser.

What are your thoughts?

EdwardMoyse:
Hiya - sorry, I missed this! Hmm. This is a good question. Maybe we can just recommend that users enable hardware acceleration and provide links explaining how? Detecting it seems a bit unreliable.

9inpachi:
@EdwardMoyse Yeah. Seems a one time popup is a better choice. Hopefully I will send a PR for this within a week.

emiliocortina:
Hi @9inpachi maybe the popup will be a little annoying to show up every time the users load the app?
I'm thinking perhaps on adding a help/user manual section inside the app where we put together some information about how to use it, and you could explain there how to enable hardware acceleration.
Let me know what you think :)

EdwardMoyse:
Yes, I think this is probably better.

9inpachi:
Note that this is a **one-time** popup which will only appears to a new user (once).
The help/user manual section is a great choice but I am not sure if many people will view it from the get go. Even if we include a button somewhere on the sides, most people will ignore it.

My proposition is that we include the hardware acceleration section in the help/user manual section and also show a **one-time** popup so that users are made aware when they first view an experiment.

emiliocortina:
@9inpachi the problem with the one-time popup is how are we going to check that the user is indeed visiting the website for the first time.
I don't know if you already have an idea for doing this, because the only thing that comes to my mind is using cookies. The thing with cookies is that for instance, here in Europe I think its use is pretty restricted by the GDPR, so I would try to avoid that.
[...]

PR #52, opened 2020-02-21
Add help panel to ATLAS experiment

9inpachi:
Addresses #35

Hi,
A help button (?) has been added to the bottom right of the ATLAS experiment which includes a section to make people aware that they can enable hardware acceleration if the rendering is too slow. Please let me know if its fine and if it should be added to other experiments so that we can create a separate help component and add it to each experiment.

**Question for @emiliocortina:** What tools do you use for the svg icons? To keep consistency, I used the other icon files and changed items inside them to generate help.svg.

Thanks. :)


EdwardMoyse:
Great! Thanks!

PR #64, opened 2020-03-12
Fix theme switching issues

9inpachi:
Hi,

There seem to have been issues switching themes.
For starters, the home page did not recognize which theme was being used and defaulted to white theme while the experiments used dark theme since they were checking which theme to use through`UIService::showUI()`.

Also the loading functions were arbitrarily setting the eventDisplay background to dark. Since UI service chooses that now, there is no need of it anymore.


emiliocortina:
Great, thank you for this!

PR #69, opened 2020-03-17
Check if cuts available in data

9inpachi:
Hi,

I have added a function `getCuts(): Cut[]` which checks if cuts are available in the collection data and returns cuts available if they exist.
Please let me know if there are any problems.

### DIFF
--- a/src/app/services/loaders/phoenix-loader.ts
+++ b/src/app/services/loaders/phoenix-loader.ts
@@ -62,55 +62,31 @@ export class PhoenixLoader implements EventDataLoader {
   }
 
   /**
-   * Returns the cuts which exist in a type of collection data.
-   * @param eventDataCollection collection data to process for available cuts.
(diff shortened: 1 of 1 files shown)


9inpachi:
That is correct, Emilio. But the current implementation does not check whether the cuts actually exist in data. Arbitrarily adding options for all the cuts makes the user believe that these cuts actually exist while the options do nothing.
We could reduce the conditional statements by only checking cuts for a single type of collection, but since each collection type has multiple collections, that also creates extra options for some collections.

Do you have something in mind that can reduce the logic or conditional statements?
I will take a look again from my side.

9inpachi:
@emiliocortina, I just realized we can simply filter to check valid values. This works great.
Not much we are adding now. Can you check?

emiliocortina:
That's perfect! I'm sorry I couldn't check it out before.

PR #71, opened 2020-03-24
Fix color change not working for collections

9inpachi:


EdwardMoyse:
Think this one is uncontroversial.

PR #72, opened 2020-03-27
Call event display service functions via console

### LINKED ISSUE #42: Add ability to add JS geometry via console in playground
In the original Phoenix and earlier, we could add our own geometry via a simple console. Would be good to have this again in the playground e.g.

var parameters = { ModuleName: "Module 2", Xdim:10., Ydim:1. , Zdim:45, NumPhiEl:64, NumZEl:10, Radius:75, MinZ:-250, MaxZ:250, TiltAngle:0.3, PhiOffset:0.0, Colour:0x00ff00, EdgeColour:0x449458  };
  window.EventDisplay.buildGeometryFromParameters(parameters);

9inpachi:
Closes #42

Hi,
Added a function to enable using event display service methods in console.
I added the ones I think would be useful. If more are to be added, let me know, I will push to this PR.

Thanks. :)

### DIFF
--- a/src/app/services/eventdisplay.service.ts
+++ b/src/app/services/eventdisplay.service.ts
@@ -2,6 +2,13 @@ import { Injectable } from '@angular/core';
 import { ThreeService } from './three.service';
 import { UIService } from './ui.service';
 import { Configuration } from './extras/configuration.model';
+import { HttpClient } from '@angular/common/http';
+
+declare global {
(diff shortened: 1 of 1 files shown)


9inpachi:
A gif of how it works:
pr-72-after

EdwardMoyse:
That’s brilliant! Thanks!

PR #76, opened 2020-04-03
Refactoring project architecture

emiliocortina:
This is a big refactoring I've been working on for quite some time.
Three Service had become quite a 'monster' with over 1.3k lines of code.
I've focused mainly on breaking down its functionality into different submodules (managers) so that it is easier to maintain.
Inside the services folder, there is a new folder called 'three' where you can find all these new managers.

As I said, this changes a lot of files, so even although these last few days I've been fixing bugs, there could be some of them left.
Hope you also find this new architecture easier to understand.


emiliocortina:
> Would be good to have #71. :)
> https://emiliocortina.github.io/phoenix/ doesn't seem to have this fix.

Thanks for pointing that out, its added now.

emiliocortina:
> Hmm. I'm not sure about this - JiveXML is _only_ an ATLAS format, and so I would argue that this is where it belongs.

They were duplicated inside the ATLAS components and inside IOOptionsDialog, and the ones that I deleted inside the ATLAS component were not called from anywhere. 

However, I am working on refactoring also the import/export menu so that it is more flexible and different for every experiment. This will mean that it can be configured inside each experiment's component, so I eventually will bring back those functions to the ATLAS component.

EdwardMoyse:
> > Hmm. I'm not sure about this - JiveXML is _only_ an ATLAS format, and so I would argue that this is where it belongs.
> 
> They were duplicated inside the ATLAS components and inside IOOptionsDialog, and the ones that I deleted inside the ATLAS component were not called from anywhere.
> 
> However, I am working on refactoring also the import/export menu so that it is more flexible and different for every experiment. This will mean that it can be configured inside each experiment's component, so I eventually will bring back those functions to the ATLAS component.

Okay, perfect.

PR #78, opened 2020-04-16
Show/hide all geometries not working for geometries loaded from GLTF

9inpachi:
Hi,

Since a group with identifier `SceneManager.GEOMETRIES_ID` is being used to group all detector geometries, geometries loaded from gltf should also be a part of this group.
The `Geometries` folder `Show` checkbox wasn't working because of this.

Thanks. :)


emiliocortina:
That's perfect, thank you!

PR #79, opened 2020-04-24
Close sections created by clipping

### LINKED ISSUE #39: Close sections created by clipping
Currently the clipping planes don't look too nice since the volumes are left open. However it looks like there are some simple hacks/workarounds which might work:

https://jsfiddle.net/wilt/qo869tb3/

9inpachi:
Fixes #39

Hi,
Found that we can edit shaders of each material before compilation which is gonna be helpful for future as well.
This changes the shaders in a way that makes clipped geometries filled. Added for both GLTF and OBJ geometries.

I was wondering if this would be better as an argument (true/false)? WDYT?

**Before:**
clipping-before

**After:**
clipping-after

### DIFF
--- a/src/app/services/three/import-manager.ts
+++ b/src/app/services/three/import-manager.ts
@@ -120,6 +120,10 @@ export class ImportManager {
                     child.material.clippingPlanes = this.clipPlanes;
                     child.material.clipIntersection = true;
                     child.material.clipShadows = false;
+                    child.material.side = DoubleSide;
+                    child.material.onBeforeCompile = (shader) => {
+                        shader.fragmentShader = shader.fragmentShader.replace('gl_FragColor = vec4( outgoingLight, diffuseColor.a );', 'if ( gl_FrontFacing ) { gl_FragColor = vec4( outgoingLight, diffuseColor.a ); } else { gl_FragColor = diffuseColor; }');
+                    };
                 }
             }
         });
--- a/src/app/services/three/import-manager.ts
+++ b/src/app/services/three/import-manager.ts
@@ -66,9 +66,6 @@ export class ImportManager {
         material2.wireframe = false;
         if (doubleSided) {
             material2.side = DoubleSide;
-            material2.onBeforeCompile = (shader) => {
-                shader.fragmentShader = shader.fragmentShader.replace('gl_FragColor = vec4( outgoingLight, diffuseColor.a );', 'if ( gl_FrontFacing ) { gl_FragColor = vec4( outgoingLight, diffuseColor.a ); } else { gl_FragColor = diffuseColor; }');
-            };
         }
 
         object3d.traverse(child => {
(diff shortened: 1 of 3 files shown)


EdwardMoyse:
That's so cool! I really didn't think this woudl work that well.

I think it would be good to be able to disable/enable it via the configuration, yes.

9inpachi:
> I think it would be good to be able to disable/enable it via the configuration, yes.

Yes, surely. Will push to this PR when done.

emiliocortina:
Hi!
Been testing this and I think it appears to not behave as expected.
It removes the shadows but it doesn't close sections, which causes visual glitches.
Screenshot 2020-04-24 at 13 46 40
Let me know if it is displayed like this on your browser.

I've also been thinking on a workaround to this but it would involve updating the 3d geometries.
I'll keep you guys updated.

9inpachi:
Hi @emiliocortina,

The results are similar to what you pointed out on my side.

image

> It removes the shadows but it doesn't close sections, which causes visual glitches.

That seems correct. I thought it was expected because of the geometries overlapping each other.

> I've also been thinking on a workaround to this but it would involve updating the 3d geometries.

The perfect solution IMO would be to redo the 3D models themselves as filled solids. Any workaround will have some bad consequence one way or the other.

9inpachi:
https://threejs.org/examples/#webgl_clipping_stencil

Found this. But looking at the code it seems not worth it considering the computation (will reduce FPS).

EdwardMoyse:
> The perfect solution IMO would be to redo the 3D models themselves as filled solids. Any workaround will have some bad consequence one way or the other.

Hi @9inpachi - AFAIK that isn't in threejs yet. I found this thread:
https://github.com/mrdoob/three.js/issues/16099
(at the end there's a ts CSG library)

Another option, which might be easier TBH, is just to build these calorimeter shapes on the fly. 

Should we accept this PR in the meantime? I think this is better than what we have right now?

9inpachi:
Hi @EdwardMoyse,
Sorry, I missed this.

CSG looks like an interesting concept. It might work in our case but it's also possible that it's use will lead to some completely different geometries. I will look into it.

> Another option, which might be easier TBH, is just to build these calorimeter shapes on the fly.

That's interesting as well. we will have to examine these geometries in detail in order to generate something reasonable.

> Should we accept this PR in the meantime? I think this is better than what we have right now?

I forgot to push some changes I did on this since this topic was on a halt. I made it configurable through the menu but there was a bug that it only worked after disabling and enabling the clipping. The visual effects don't take place right away for some reason even if we set the material to update. Mostly because shaders aren't built that often and we can't manually build them.

EdwardMoyse:
I'll leave this one for the moment.

sponce:
Is this definitively gone ? I'm asked by the LHCb outreach people whether there is a way to get precisely this functionality.

EdwardMoyse:
Probably we should close this @9inpachi

9inpachi:
> Probably we should close this @9inpachi

Yes. I don't think it's a stable enough implementation.

PR #80, opened 2020-05-06
Add info panel

### LINKED ISSUE #60: Add info panel
It would be good to have an info panel so we can output information e.g. just loaded event#12345 from run#12222. Or, clicked on Track # 39 with pT=x.

9inpachi:
Closes #60 

### Description
I added an info logger as a service. And added info logging for some functions such as loading OBJ geometry, loading GLTF geometry, exporting the scene, info about the event and run etc.

### Usage
We can instantiate the info logger through the constructor or passing it to the called function if the constructor is not a viable option.
Then we just use `this.infoLogger.add(text: string)` to add what we want to display on the info panel.

### Screen Capture
Here is how it looks:
info-panel


EdwardMoyse:
I'd be happy to accept this. Any objections anyone?

EdwardMoyse:
Hmm. I probably should have accepted this before #81 . Sorry! 

@9inpachi can you take a look? I don't want to make more of a mess.

9inpachi:
This is fine. I forgot to mention it but this was to updated after the #81 was accepted.
I will update and let you know when it's safe to merge.

EdwardMoyse:
Okay thanks!

9inpachi:
Should be good now.

EdwardMoyse:
Oops. "This branch cannot be rebased due to conflicts"

9inpachi:
This should fix it.

Issue #57, opened 2020-03-11
Make overlay more functional


EdwardMoyse:
I'm thinking that it would be good to be able to resize it and switch it from orthographic/perspective independently. Maybe on mouse over, some controls could appear - a pin to fix it, and cube/perspective cube for ortho/persp?

9inpachi:
Ed, by
> a pin to fix it

Do you mean to fix the overlay view so it doesn't move or zoom with the main view controls?

9inpachi:
Fixed in #81 and 30e568fdec28511491f28e73c7737e4af189e407.

PR #81, opened 2020-05-12
Improvement of overlay components

9inpachi:
Hi,

### Description
Improved the overlay components and created a parent overlay component to make overlays consistent.

### Added
* Option for expand/collapse
* Ability to resize
* Drag by moving the overlay header
* Added icon to header
* Improved and consistent design

### Screen Capture
wip-overlays


EdwardMoyse:
This is great! Any objections to accepting?

9inpachi:
**Updated Screen Capture:**

wip-overlays-1

As in #57, I added a button to switch overlay view from the overlay window. About pinning an overlay, as mentioned in the issue, I am not sure if it is applicable for the overlays. My understanding of the pin is that it will pin the overlay, disabling drag (and may be expand/collapse also) but that doesn't seem useful. WDYT?
Let me know if you have other ideas.

EdwardMoyse:
Thanks for this @9inpachi  - this is very nice!

I don't think pin is terribly useful. If we don't want it to move, we will just not drag it!

PR #84, opened 2020-05-30
Add toggle for wireframing geometries

9inpachi:
Hi,

Added a toggle for wireframing detector geometry and decreasing it's opacity for a better event view. Here's how it looks:
image


EdwardMoyse:
Very nice!

PR #95, opened 2020-06-11
Set the pixelRatio to be that of the device. 

EdwardMoyse:
Should make output look better.

Before:
image

After:
image


9inpachi:
It looks a little better but normally using antialias or something like the FXAAShader (https://threejs.org/examples/?q=postpr#webgl_postprocessing_fxaa) we shouldn't get pixelated lines at all.

PR #97, opened 2020-06-15
Improvement of geometries

9inpachi:
Hi,

These are some of the improvements to geometries and the overall scene.

**What's Added:**
* Use glTF geometries for ATLAS
* Add option to glTF loader for scaling geometries
* Use the same side (DoubleSide/FrontSide) for geometries as given in the glTF file
* Improve geometries by updating materials and some other stuff
* Increase camera frustum near plane which fixes all of the shader bugs with normal geometries
* Add option for using a single directional light that follows the camera (a little improved performance)
* Add option for using four directional lights that illuminate the entire scene

**Screenshots:**

**Before:**
image

**After:**
image


src/app/services/three.service.ts
```diff
@@ -251,25 +250,6 @@ export class ThreeService {
     this.importManager.parseGLTFGeometry(geometry, callback);
   }
 
-  /**
-   * Load geometries from JSON.
-   * @param name Name of the geometries or group of geometries.
-   * @param json JSON containing all the geometries.
-   */
-  public loadJSONGeometry(name: string, json: string | object) {
```
EdwardMoyse: Hi @9inpachi can you remind me why you removed this? I think you said that it seems not to be so well supported, but it is in current versions of threejs I think?

9inpachi: It more or less doesn't work. I tried to import some JSON but it threw errors.
Would you like me to give it a try again? But this time with JSON exported from Phoenix itself.

9inpachi: So I tested it out and exporting and importing using the newer version works.
I am re-adding this method through the import manager. Should I add methods for export as well?

EdwardMoyse:
I think export isn’t necessary. This isn’t the primary role of Phoenix.

Cheers,

Ed

9inpachi:
This should be good now. I will add the PR for metadata once this is merged.

PR #98, opened 2020-06-16
Selection of 3D object from collections info panel

### LINKED ISSUE #73: Line width issue for tracks
Hi all,

The function `getTrack(trackParams: any): Object3D` currently returns an object of `THREE.Line`.
The problem is that `THREE.LineBasicMaterial` has known bugs for the linewidth property which does not work on windows and on chrome for MacOS. See [THIS](https://github.com/mrdoob/three.js/issues/14627) for details.

I was working on a control for changing line width of tracks and came across this problem. The solution is to use either Line2 or THREE.MeshLine.

**Using THREE.MeshLine**
I have a solution using MeshLine [here](https://github.com/9inpachi/phoenix/tree/testingMeshLine) but the problem with it is that it scales with camera zoom since it is a **Mesh**. We can scale it for every render (`animate()`) we do but that is too expensive in terms of computation.

Here is what it looks like:

tracks-meshline-after

**Using Line2**
I also have a solution ready using `Line2` [here](https://github.com/9inpachi/phoenix/tree/line2tracks). `Line2` is not part of three.js core and we have to use scripts in `three/examples/jsm/lines/`. `Line2` does not have any scaling issues so I think this is the perfect choice.

Here is what it looks like using `Line2`
 
line2tracks-after-low

WHAT ARE YOUR THOUGHTS?
This is surely going to make tracks selection MUCH SIMPLER.

### LINKED ISSUE #93: Connect collections information and 3D view
If you click on a cluster, track etc, and the collections view is open, that object should be highlighted there. And vice versa. 

It would also be great to have an option (in the collections panel?) to zoom to that object.

9inpachi:
Fixes #93 

Hi,

## Description
Updated the collections info panel so that selection in the scene will highlight the event data items in the collections info panel and users can also highlight or move to objects in the scene from the collections info panel.

## Added
* Highlight event data in collections info panel on selection of event object from the scene
* Highlight event data in the scene on selection of event data from the collections info panel
* Ability to highlight and move camera to object from the collections info panel
* Works for all event data including Tracks, Muons, Hits and Clusters
* Two different options "Move camera to object" and "Highlight object" for highlighting event data from collections info panel
* Some fixes to the collections info panel
* A combination of `Tube` and `Line` for better selection of Tracks (Fixes #73)
* Use yellow color for Tracks if the theme is dark (should I remove this or is it fine? [CODE](https://github.com/9inpachi/phoenix/blob/wip-collections/src/app/services/loaders/objects/phoenix-objects.ts#L36))

## Screen Capture
wip-collections-final

### DIFF
--- a/src/app/components/ui-menu/collections-info/collections-info-overlay/collections-info-overlay.component.html
+++ b/src/app/components/ui-menu/collections-info/collections-info-overlay/collections-info-overlay.component.html
@@ -1,6 +1,7 @@
 <!-- Show objects -->
 <app-overlay title="Collections Info" icon="info" [resizable]="true" [active]="showObjectsInfo">
-  <div class="collectionsInfo m-2">
+  <!-- Using ngIf to remove it from DOM since this panel requires heavy computing -->
+  <div *ngIf="showObjectsInfo" class="collectionsInfo m-2">
     <div class="collectionSelector mb-2 d-flex align-items-center" *ngIf="collections!=null">
       <span>Choose a collection: </span>
       <div class="eventSelector">
--- a/src/app/components/ui-menu/collections-info/collections-info-overlay/collections-info-overlay.component.html
+++ b/src/app/components/ui-menu/collections-info/collections-info-overlay/collections-info-overlay.component.html
@@ -29,12 +29,14 @@
             <td>#{{i}}</td>
             <td>
               <div *ngIf="object.uuid" class="row justify-content-center object-select">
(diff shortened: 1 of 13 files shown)


EdwardMoyse:
> Use yellow color for Tracks if the theme is dark (should I remove this or is it fine? CODE)

Hmm. I can see why you'd want to do this, but I'm not sure about hardcoding changes like this. I would vote to remove it.

PR #99, opened 2020-06-17
Dynamically get metadata for experiment info

9inpachi:
Hi,

## Description
This adds dynamic metadata to display in the experiment info section.

## Added
* Get run number, event number and LS from event data
* Add support for optionally using an object with label and value (label is bold)
* Tagline for experiment

## Screenshot
image


EdwardMoyse:
This is so cool!

PR #107, opened 2020-07-02
Scalable Jets and fix depthTest

### LINKED ISSUE #103: Make it possible to scale jets / energies in general
This should be something the user can tweak (in options, which we should also add) but also we should consider setting a sensible autoconfiguration ... the jets should never extend too far.

image

9inpachi:
Closes #103 

Hi,

## Description

This makes the Jets scalable and also allows to extend UI menu options for each object type which can be used with any loader (by using the `PhoenixLoader.addObjectType` method from a child class).

I have been trying out options to scale Jets since yesterday but removing and recreating the Jets with increased energy wasn't working for large number of Jets. So after a lot of tries I came up with a minimal method which I think is the best.

## Added
* Fixed depthTest not working correctly
* Removed custom synchronous function to loop through children for changing depthTest (I am not sure why but `Object3D.traverse` seems to work now - while depthTest only worked with `updateChildrenDepthTest` function earlier)
* Made Jets scalable by increasing their scale and changing their position
* Ability to scale Jets from 1% to 200%
* Increased maximum Jet length to 3000 as it looks like a better option
* **Mechanism to add exclusive options to each event data type folder of dat.GUI through an optional callback passed to `PhoenixLoader.addObjectType`**

## Screen Capture
wip-scalable-jets

### DIFF
--- a/src/app/services/loaders/objects/phoenix-objects.ts
+++ b/src/app/services/loaders/objects/phoenix-objects.ts
@@ -97,8 +97,8 @@ export class PhoenixObjects {
     // Jet energy parameter can either be 'energy' or 'et'
     let length = (jetParams.energy ? jetParams.energy : jetParams.et) * 0.2;
     // Ugh - We don't want the Jets to go out of the event display
-    if (length > 2000) {
-      length = 2000;
+    if (length > 3000) {
+      length = 3000;
     }
     const width = length * 0.1;
 
--- a/src/app/services/loaders/phoenix-loader.ts
+++ b/src/app/services/loaders/phoenix-loader.ts
@@ -123,7 +123,15 @@ export class PhoenixLoader implements EventDataLoader {
         new Cut('energy', 2000, 10000)
       ];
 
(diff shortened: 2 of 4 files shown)


src/app/services/loaders/objects/phoenix-objects.ts
```diff
@@ -97,8 +97,8 @@ export class PhoenixObjects {
     // Jet energy parameter can either be 'energy' or 'et'
     let length = (jetParams.energy ? jetParams.energy : jetParams.et) * 0.2;
     // Ugh - We don't want the Jets to go out of the event display
-    if (length > 2000) {
-      length = 2000;
+    if (length > 3000) {
```
EdwardMoyse, line 100: Fine for now, but I think this is something we should make configurable (the appropriate maximum length will depend a lot on the experiment)

9inpachi: I agree. May be we can apply some sort of normalization.

It's also possible to check this directly in the loaders and change the energies from there.

PR #114, opened 2020-07-24
Option to load geometry on demand

### LINKED ISSUE #106: Add option to load geometry on demand
Since geometry can be large, we should make it possible to conditionally load the geometry. i.e. have a placeholder in the menu, but only actually load the geometry when clicked on.

9inpachi:
Closes #106 

Hi,

## Description

This adds option for making the geometry initially invisible through the geometry loading functions.
**NOTE:** @EdwardMoyse let me know which geometries from ATLAS are to be made initially invisible. I will push to this PR once I know.

## Added

* Added param `initiallyVisible: boolean = true` to `loadOBJGeometry`, `parseOBJGeometry`,  `loadGLTFGeometry` and `loadJSONGeometry` functions of `EventdisplayService` for making geometry invisible (initiallyVisible = false) by default
* Fix preset views not loading (set the configuration property of `UIService` through `showUI` instead of `showDatGUIMenu` since dat.GUI menu is optional now)

## Screen Capture

(With "Toroids" and "Tile Cal" initially invisible)
fix-geometry-on-demand

### DIFF
--- a/src/app/sections/atlas/atlas.component.ts
+++ b/src/app/sections/atlas/atlas.component.ts
@@ -31,14 +31,14 @@ export class AtlasComponent implements OnInit {
     this.eventDisplay.init(configuration);
     this.http.get('assets/files/event_data/atlaseventdump2.json')
       .subscribe((res: any) => this.eventDisplay.parsePhoenixEvents(res));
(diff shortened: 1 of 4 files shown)


EdwardMoyse:
Hi @9inpachi - this is great! I think for the moment, I would make Toroids and the calorimeters invisible by default

PR #117, opened 2020-08-01
JSROOT Support

9inpachi:
Hi,

## Description

This adds JSROOT support to Phoenix with ability to load JSROOT scripts dynamically through a script loader, support for loading JSROOT's `.json.gz` and `.root` geometries and support for JSROOT event loader for loading event data from `.root` files.

Went through the JSROOT core code and edited the code (changed script paths) for proper integration. Deeply analyzed the JSROOT code including the `JSRootCore` and `JSRootPainter` and other files for getting and processing the event data from `.root` files.

## Added

* A generalized `ScriptLoader` for dynamically loading js script files without duplication
* Ability to dynamically load JSROOT script files through `ScriptLoader`
* Function for loading geometry object from `.root` and `.json.gz` files through the event display service
* A working JSROOT event loader
* Support for loading event data from `.root` file objects
* Loading multiple event data objects (tracks, hits etc.) from a `.root` file
* `JSRootEventLoader` for processing `.root` files to get `TEveTrack`s, `TGeoTrack`s and `TEvePointSet`s (Hits)
* Use ROOT JSON geometry for the CMS experiment (more detailed - **let me know if this is to be reverted back to the old geometry**)

Misc

* Don't sort track positions (`pos`) because it leads to weird behavior for extrapolated tracks whose curve might return back
* Fixed all the failing tests
* Add parameter for making the geometry's side DoubleSide with `loadJSONGeometry` function

## Usage (Example Code Snippets)

Loading `.json.gz` and `.root` geometries.

```ts
ScriptLoader.loadJSRootScripts((JSROOT: any) => {
  this.eventDisplay.loadRootJSONGeometry(JSROOT, 'https://root.cern/js/files/geom/cms.json.gz',
    'CMS Detector', 10, true);
  this.eventDisplay.loadRootGeometry(JSROOT, 'https://root.cern/js/files/geom/atlas2.root',
    'atlas', 'ATLAS Detector', 10, true);
});
```

Using `JSRootEventLoader`.

```ts
ScriptLoader.loadJSRootScripts((JSROOT: any) => {
  const jsrootEventLoader = new JSRootEventLoader(JSROOT, 'https://root.cern/js/files/geom/tracks_hits.root');
  jsrootEventLoader.getEventData(['tracks;1', 'hits;1'], (eventData: any) => {
    this.eventDisplay.buildEventDataFromJSON(eventData);
  });
});
```

## Screenshots

ROOT JSON geometry.

image

JSROOT event loader's `TEveTrack`s.

image

JSROOT event loader's `TEveTrack`s and `TEvePointSet`s (Hits).

image

**Note:** @EdwardMoyse Can you please share a `.root` file with `TGeoTrack` objects? I couldn't find it in the JSROOT files [here](https://root.cern/js/files/).


EdwardMoyse:
This looks great!

EdwardMoyse:
Hi @9inpachi  - this adds jsroot into Phoenix (which I think is okay from the point of view of licences) but I gather you had to modify parts of it? Does this mean that if a new jsroot version comes out, these changes would need to be made again? If so, perhaps it would be good to add a README explaining this.  Otherwise, this looks great!

9inpachi:
Hi @EdwardMoyse,

Yes, indeed I changed some parts of it mostly related to the script directories for scripts being loaded within JSRoot files. I updated it later on to not depend on the internal script loading but I didn't revert the things I changed. I will try reverting them back and see if our `ScriptLoader` works fine with the raw version of JSRoot. And then if there's a need for some README docs, I'll be sure to write them.

9inpachi:
So I checked and it indeed doesn't work without changing the script paths from within JSRoot core files. I will write down the procedure here for now and formally document it later along with some other things.

**TL;DR** - For updating a JSRoot version, in the `src/assets/jsroot` directory of Phoenix, we just need to replace all `script/` occurrences with `assets/jsroot/`.

What happens is that before most function calls, JSRoot checks using `JSROOT.AssertPrequisites` if the scripts are loaded or not. Unfortunately, JSRoot maintains it's own list of scripts loaded through any of the natively supported methods (requirejs, node etc.) so the scripts we dynamically load using the Phoenix's `ScriptLoader` are not accounted for and the code doesn't go through before actually loading the scripts through JSRoot code. There are a couple of other complex reasons (like loading scripts dynamically and completely changing some JSROOT functions like JSROOT.OpenFile through the loaded code) within the library for why it doesn't work but I don't have a 100% understanding of them.

PR #119, opened 2020-08-16
Keyboard controls

9inpachi:
**Note:** To be merged after PR 118 

Hi,

## Description

This adds keyboard controls for common event display and three service functions.

## Added

* Keyboard controls for
  * Toggling theme (`shift + t`)
  * Going to preset view (`shift + 1..9` - 1 to 9 is the index of preset view with 1 being the first preset view)
  * Autorotate (`shift + r`)
  * Zoom in (`shift + "+"`)
  * Zoom out (`shift + "-"`)
  * 180° Clipping (`shift + c`)
  * Toggling view (`shift + v`)


EdwardMoyse:
Great!

Did you add documentation anywhere for what the keys do? If not, we should add this (but it can be in a follow up PR to just get this in)

9inpachi:
> Did you add documentation anywhere for what the keys do? If not, we should add this (but it can be in a follow up PR to just get this in)

No. I don't think there is a common place where we can put this. I was thinking of creating a detailed user manual mostly covering how to use Phoenix's event display and other functionalities. I think I will be able to write it sometime by the end of this year. But not during GSoC because I am focusing on the code for now. ^_^

PR #121, opened 2020-08-19
RungeKutta

EdwardMoyse:
This is a RungeKutta implementation, based on that of @asalzburger here:
https://github.com/asalzburger/rungekutta
translated to TypeScript by @9inpachi 

Currently it is not being used, but in this branch we did earlier some tests to check the performance on ATLAS tracks


9inpachi:
WONDERFUL Ed.

Thanks for this. I will look into the conflicts and merge it once the other PRs are merged. This has the most conflicts since it's old -  so I will need to see in details since this branch has some other fixes as well.

EdwardMoyse:
Hi @9inpachi - I tried to rebase, but please have a look ... there were a few changes I was unsure about.

PR #130, opened 2020-08-28
Fix old geometry

EdwardMoyse:
@9inpachi for me locally this seems to fix the material becoming black with r120.


EdwardMoyse:
Oops. But I have dragged in your VR commits.

9inpachi:
Great! The new geometries are working like a charm.
I downgraded my three.js to r117 yesterday but the geometries were still showing up black.

Cherry-picked in d426855e210c43f4926c63d833d489539c068179.

PR #138, opened 2020-09-09
[WIP] Jive xml improvements

EdwardMoyse:
This adds some new types of objects to Phoenix (not all can be displayed yet) and also adds the ability to truncate the Track extrapolation at a Volume. For example, for ATLAS it doesn't really make sense to extrapolate tracks past the entrance to the calorimeter.

This is not very elegant TypeScript - @9inpachi I'd appreciate a code review! I bet there are better ways to do a lot of this, especially the changes to the RK, and the conditional checks on the JiveXML contents (some tracks have some variables, which are missing on others).

There is one *big* problem at the moment - opening some jiveXML is very very slow (it does eventually complete) because the tracks are very low transverse momenta and so get extrapolated a long way. I think the path length and step size could help, but also it would be good to not try to propagate tracks with too low momenta.

### DIFF
--- a/src/app/services/extras/runge-kutta.ts
+++ b/src/app/services/extras/runge-kutta.ts
@@ -85,11 +85,12 @@ export class RungeKutta {
    * @param q Charge.
    * @param mss Max step size.
    * @param plength Path length.
+   * @param inbounds Function which returns true until the passed position is out of bounds, when it returns false.
    * @returns An array containing position and direction at that position calculated
    * through the Runge-Kutta steps.
    */
   static propagate(startPos: Vector3, startDir: Vector3, p: number, q: number,
-    mss: number = -1, plength: number = 1000): { pos: Vector3, dir: Vector3 }[] {
+    mss: number = -1, plength: number = 1000, inbounds: any): { pos: Vector3, dir: Vector3 }[] {
     let rkState: State = new State();
     rkState.pos = startPos;
     rkState.dir = startDir;
@@ -99,7 +100,7 @@ export class RungeKutta {
 
     let result: { pos: Vector3, dir: Vector3 }[] = [];
 
-    while (rkState.pathLength < plength) {
+    while (rkState.pathLength < plength ) {
       rkState.pathLength += RungeKutta.step(rkState);
       // Cloning state to avoid using the reference
       let copiedState = JSON.parse(JSON.stringify(rkState));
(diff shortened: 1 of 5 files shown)


9inpachi:
> There is one big problem at the moment - opening some jiveXML is very very slow (it does eventually complete) because the tracks are very low transverse momenta and so get extrapolated a long way.

I think we should implement a progressive loader. At least to tell the users that something is being processed. Not sure if we should have a common one through a service.

src/app/services/extras/runge-kutta.ts
```diff
@@ -85,11 +85,12 @@ export class RungeKutta {
    * @param q Charge.
    * @param mss Max step size.
    * @param plength Path length.
+   * @param inbounds Function which returns true until the passed position is out of bounds, when it returns false.
    * @returns An array containing position and direction at that position calculated
    * through the Runge-Kutta steps.
    */
   static propagate(startPos: Vector3, startDir: Vector3, p: number, q: number,
-    mss: number = -1, plength: number = 1000): { pos: Vector3, dir: Vector3 }[] {
+    mss: number = -1, plength: number = 1000, inbounds: any): { pos: Vector3, dir: Vector3 }[] {
```
9inpachi, line 93: We can define the type here to be `inbounds: (pos: Vector3) => boolean`.

src/app/services/extras/runge-kutta.ts
```diff
@@ -99,14 +100,20 @@ export class RungeKutta {
...
+    while (rkState.pathLength < plength ) {
       rkState.pathLength += RungeKutta.step(rkState);
       // Cloning state to avoid using the reference
       let copiedState = JSON.parse(JSON.stringify(rkState));
       result.push({
         pos: copiedState.pos,
         dir: copiedState.dir
       });
+      // Ugly
+      if ( inbounds(copiedState.pos) != true){
```
9inpachi, line 112: If we have the parameter above typed correctly (so the compiler knows that it must return boolean), I think it's safe to use just the function - `if (inbounds(copiedState.pos))`.

src/app/services/helpers/rk-helper.ts
```diff
@@ -25,9 +25,11 @@ export class RKHelper {
   /**
    * Extrapolate tracks using RungeKutta propagator.
    * @param track Track which is to be extrapolated.
+   * @param inbounds Function which returns true until the passed position is out of bounds, when it returns false. Default just always returns true.
    * @returns An array of track positions.
+
    */
-  public static extrapolateTrackPositions(track: { dparams: any }): any {
+  public static extrapolateTrackPositions(track: { dparams: any },  inbounds: any=function() {return true;}): any {
```
9inpachi, line 32: The type for `inbounds` here can become - `inbounds: (pos: Vector3) => boolean = (new Vector3()) => true`.

9inpachi:
Ed - I think it's fine to merge this without worrying much. I do have to integrate it with the angular independent repo I am working on - so I can refactor the code myself then.

PR #141, opened 2020-09-18
Jive xml improvements

EdwardMoyse:
Added "pretty print" functionality, which let's developers chose which information to show on object selection:
image

Read some more variables from JiveXML and add protections against duplicate and problematic collections.

Fixed issue with MeV vs GeV for pT.

### DIFF
--- a/packages/phoenix-event-display/src/loaders/jivexml-loader.ts
+++ b/packages/phoenix-event-display/src/loaders/jivexml-loader.ts
@@ -84,7 +84,12 @@ export class JiveXMLLoader extends PhoenixLoader {
    */
   private getNumberArrayFromHTML(collection: Element, key: any) {
     // console.log(collection);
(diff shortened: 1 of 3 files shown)


EdwardMoyse:
@9inpachi the error isn't related to this PR so I'm just going to accept it, okay?

`Error: Command failed: compodoc -p compodoc.conf.json -n "Phoenix API Docs" -d ../../docs/api-docs

/bin/sh: 1: compodoc: not found`

9inpachi:
@EdwardMoyse I was aware of the error. Just pushed a fix with a couple other things improved.

About the pretty print functionality. I think we should aim towards a more generalized approach. Since this only works with data that has dParams. So tracks from some other source, for example CMS, don't get pretty printed parameters.

EdwardMoyse:
Ok. That’s a good point! I should look at LHCb tracks too.

9inpachi:
I have an approach in mind (somewhat rough), I will try implementing it tomorrow if I get the time.

Issue #133, opened 2020-09-01
Cuts aren't initialised with the correct values


EdwardMoyse:
They start as zero - they shoudl show the existing values.

![image](https://user-images.githubusercontent.com/6764617/91867095-8382c100-ec73-11ea-9474-06881bcc26cd.png)

9inpachi:
@EdwardMoyse just wanted to confirm. Is that how it worked before?

EdwardMoyse:
I'm not sure this ever worked right, no. I just noticed that the track colour, and the cuts, are not set in the GUI as they are in the code.

PR #145, opened 2020-10-02
Range slider in phoenix menu and other important improvements

### LINKED ISSUE #142: Add ranged cuts
This would be nice in general, but especially for the clipping cuts, where currently we can only clip from 0 - 180. What would be better is to be able to clip to arbitrary ranges e.g. 200-270 degrees.

What I mean is something like:

image

https://www.bestjquery.com/?wxmleVah

### LINKED ISSUE #144: Cuts not working after implementing pretty params
It seems using the pretty params directly as objects' `userData` causes cuts to not work.

Possible options are to

* Either put pretty params in a separate property of `userData` object
* Or use the new pretty param symbols as cut fields

Not sure which is more stable. For now I think we should go with putting pretty params inside a `userData` object property since it's something we process through the event display and not a raw part of event data.

9inpachi:
Closes #144
Closes #142
And it might be safe for #133 to close with this as well.

Hi,

## Description

This adds range slider to phoenix menu item configuration. The new range slider uses [ng5-slider](https://github.com/angular-slider/ng5-slider) with custom styling. This also includes fixes and improvements for pretty parameters and cuts.

## Added

* Range slider for phoenix menu configuration and use range slider for cuts
* Fix cuts not working after implementing pretty parameters
* Use pretty parameters only where required and don't change the actual parameters in event data
* Use pretty parameters for cuts
* Move pretty parameters to a separate helper class (having it in `PhoenixObjects` isn't really optimal)
* Improve and simplify code for getting pretty parameters
* Some fixes and improvements for cuts

## Screenshots

image


9inpachi:
Closes #144, #142
And it might be safe to close #133 with this as well.

EdwardMoyse:
Wow - this looks GREAT! Thanks @9inpachi !

PR #148, opened 2020-10-11
Save and load Phoenix menu state

9inpachi:
Hi,

## Description

This adds feature to save and load configuration aka state of the Phoenix menu with effects to the scene (like event data color, geometry opacity etc.). This works well for all configurations in the Phoenix menu. It saves the Phoenix menu state as a `json` file which can later be loaded. The state can also be loaded directly from code (using `PhoenixMenuNode.loadStateFromJSON(json: object } string)`.

It works REALLY WELL and will be very useful. I didn't think it would turn out this good.

The PR includes some other fixes and improvements as well.

## Added

* Save state of the Phoenix menu as JSON  
  * Save state of the node like the toggle state, all configuration options (like color and slider values) of the node and the state of all its children
* Process and load the state of the Phoenix menu from JSON  
  * Load state of the node like the toggle state, process all configuration options (like color and slider values) and apply them to the objects in scene and load state of all of its children
* Add options to save and load state to the root Phoenix menu node

Misc

* Set the default event data collection color on load in the Phoenix menu
* Set the default geometry color on load in the Phoenix menu
* Fix CI and use lerna based commands in CI

## Screen Captures

**Setting and saving the Phoenix menu state**

wip-config-save

**Loading the Phoenix menu state in a fresh scene**

The options we set in the upper screen capture will all be loaded on loading the state.

wip-config-load

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,17 +11,14 @@ addons:
 cache:
   directories:
     - ./packages/phoenix-app/node_modules
-
-before_install:
-  - cd ./packages/phoenix-event-display
-  - npm link
-  - cd ../phoenix-app
+    - ./packages/phoenix-event-display/node_modules
 
 install:
-  - npm install
-  - npm link "phoenix-event-display"
+  - npm install lerna --global
+  - npm run install:dependencies
 
 script:
+  - cd ./packages/phoenix-app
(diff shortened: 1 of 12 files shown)


EdwardMoyse (approved):
This looks fabulous Fawad! Thank you!

PR #153, opened 2020-10-21
Add default view

EdwardMoyse:
This adds a defaultView to the configuration, and moves the ATLAS defaultView to a position which is a bit more performant, and perhaps less confusing.


9inpachi:
Great. I planned this week to add camera view with the save / load state feature of Phoenix menu.

9inpachi:
And Ed, before pushing code, it'd be great if you could format the code files using VSCode. You can either "Right Click -> Format Document" or use "Shift + Alt + F" to format the document or file. Just for code style consistency.

EdwardMoyse:
It's not formatted? It looks fine to me - even in github. Can you point to what looks wrong?

9inpachi:
Just nitpicking from my side. The equals don't have space before and after and things like that. The auto format is really helpful if something skips our eyes.

image

EdwardMoyse:
Ah - great! I didn't realise that this was enforced too (I though it was just indents). Will do so in future.

PR #158, opened 2020-10-22
feat(app): load event from URL query params in ATLAS

9inpachi:
ONLY FOR ATLAS

This adds the ability to load an event by specifying a file through the URL.

**Cases:**

1. If both `file` AND `type` parameters are not defined, the default event (JiveXML) will be loaded
1. If both `file` AND `type` parameters are defined, the event path in the URL will be used
1. Both `file` AND `type` are required for loading an event through the URL. If neither  defined, the default event will be used

**Example URLs:**

* http://localhost:4200/atlas?file=assets/files/FJiveXML/JiveXML_359310_3066561649.xml&type=jivexml
* http://localhost:4200/atlas?file=assets/files/event_data/atlaseventdump2.json&type=json
* http://localhost:4200/atlas (default event)


EdwardMoyse:
I'm so glad you did this instead of me! This is much more elegant! ;-)

PR #160, opened 2020-10-23
feat(app): Add a deploy:atlas script. Use hashed routes.

EdwardMoyse:
This is to enable us to deploy to a server at CERN and have URL file and config work.

### DIFF
--- a/packages/phoenix-app/README.md
+++ b/packages/phoenix-app/README.md
@@ -49,3 +49,9 @@ npm run start
 ```
 
 Now any changes in the [phoenix-event-display](https://www.npmjs.com/package/phoenix-event-display) package ([source](https://github.com/HSF/phoenix/tree/master/packages/phoenix-event-display)) will build the package and the Angular app will pick up the changes made to the event display.
+
+### Deploy the application
+In order to make a version of Phoenix which can be deployed to your server, run the following command:
+```sh
+npm run deploy:web -- --base-href "/hsf-phoenix/atlas/"
+```
--- a/packages/phoenix-app/README.md
+++ b/packages/phoenix-app/README.md
@@ -51,7 +51,12 @@ npm run start
 Now any changes in the [phoenix-event-display](https://www.npmjs.com/package/phoenix-event-display) package ([source](https://github.com/HSF/phoenix/tree/master/packages/phoenix-event-display)) will build the package and the Angular app will pick up the changes made to the event display.
 
 ### Deploy the application
-In order to make a version of Phoenix which can be deployed to your server, run the following command:
+In order to make a version of Phoenix which can be deployed to your server, from the `packages/phoenix-app` directory, run the following command:
 ```sh
-npm run deploy:web -- --base-href "/hsf-phoenix/atlas/"
+npm run deploy:web
 ```
+You can then copy it to your server e.g. with:
+```sh
+rsync -avz docs/ your-server.net:path/to/website
+```
+
--- a/packages/phoenix-app/package.json
+++ b/packages/phoenix-app/package.json
@@ -11,8 +11,8 @@
   },
   "scripts": {
     "ng": "ng",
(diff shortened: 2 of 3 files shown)


packages/phoenix-app/package.json
```diff
@@ -11,16 +11,17 @@
   },
   "scripts": {
     "ng": "ng",
-    "start": "ng serve",
-    "start:ssl": "ng serve --ssl",
+    "start": "ng serve --ssl --host=0.0.0.0",
```
9inpachi, line 14: I think we don't need the ssl and host options in this. We mostly want to use that when testing VR so we can have the `start:ssl` script for that.

If we want the `phoenix-event-display` to start with the app with lerna, we can add a duplicate `start:ssl` script in `phoenix-event-display` which will just be `"start:ssl": "npm run start"`.

EdwardMoyse: Oops yes this is a mistake and shouldn't have been committed. Thanks!

packages/phoenix-app/package.json
```diff
@@ -12,15 +12,16 @@
...
     "build": "ng build",
     "test": "ng test",
     "lint": "ng lint",
     "e2e": "ng e2e",
     "deploy": "mv ../../docs/api-docs ../../api-docs && ng build --prod --output-path ../../docs --base-href \"/phoenix/\" && cp ../../docs/index.html ../../docs/404.html && mv ../../api-docs ../../docs/api-docs",
     "test:coverage": "rm -rf ./coverage && ng test --no-watch --code-coverage",
     "coveralls": "cat ./coverage/phoenix/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
-    "version": "npm run deploy"
+    "version": "npm run deploy",
+    "deploy:atlas": "ng build --prod --output-path ./atlas --base-href \"/hsf-phoenix/atlas/\" && cp ./docs/index.html ./docs/404.html"
```
9inpachi, line 24: I think it's better if we generalize this script for deploying anywhere.
We can make this:

```
"deploy:web": "ng build --prod --output-path ./docs",
"postdeploy:atlas": "cp ./docs/index.html ./docs/404.html"
```

And then we can define the `--base-href` when calling the script:

```
npm run deploy:web -- --base-href "//hsf-phoenix/atlas/"
```

(You might need a single `/` instead of `//`. Using `//` at start here because git bash somehow converts to a full windows path otherwise)

9inpachi (changes requested):

packages/phoenix-app/package.json
9inpachi: This single script works for any path (I checked on multiple paths on Apache). We can just run this command and put it anywhere on the server and it will work.

```
"deploy:web": "ng build --prod --output-path ./docs --base-href \"./\" && cp ./docs/index.html ./docs/404.html"
```

packages/phoenix-app/README.md
```diff
@@ -49,3 +49,9 @@ npm run start
 ```
 
 Now any changes in the [phoenix-event-display](https://www.npmjs.com/package/phoenix-event-display) package ([source](https://github.com/HSF/phoenix/tree/master/packages/phoenix-event-display)) will build the package and the Angular app will pick up the changes made to the event display.
+
+### Deploy the application
+In order to make a version of Phoenix which can be deployed to your server, run the following command:
+```sh
+npm run deploy:web -- --base-href "/hsf-phoenix/atlas/"
```
9inpachi, line 56: Sorry I think I didn't mention this in the second comment of the last review. When using `--base-href \"./\"` in the `deploy:web` npm script we don't need to specify the `--base-href` again. Only running `npm run deploy:web` will build the app we can put anywhere on the server.

9inpachi (changes requested):

9inpachi (approved):
This is good to merge.

PR #162, opened 2020-10-27
Implicitly load event data from URL with event display

9inpachi:
This adds support to `phoenix-event-display` for loading an event by default if the `file` and `type` are defined in the URL query parameters.

We can specify a default fallback event to load if the parameters in the URL are not defined.

```
configuration.defaultEventFile = {
  eventFile: 'path/to/file.json',
  eventType: 'json'
}
```

In case the default fallback event and the URL parameters both are not defined then no event will be loaded onto the event display.


9inpachi:
@EdwardMoyse I think it's better if we discuss this a little. I was skeptical about this when I implemented it but didn't know how to put it forward.

So I think that implicitly having the event display load an event without any configuration can cause problems - mostly for custom loaders. Let us take the CMS experiment for example which uses the `CMSLoader`. The behavior introduced by this PR will load a particular event specified in the URL even if it is the CMS experiment. Not to mention that if we load an event manually in a section like we do in the CMS section, it can cause problems because if we also have an event specified in the URL, both the URL event and the manually defined in the section will load simultaneously causing problems. On the other hand, if we have a configuration option to enable/disable this feature we can have it disabled by default and have a single configuration option decide if it should be implicitly enabled or not.

WDYT?

I still don't have a firm grasp on it myself - so I don't have any solid workaround besides a configuration option.

EdwardMoyse:
I think since we already have the IO dialog which permits loading of incompatible event data, that this isn't worse.

What might be good is to have a configuration option which specifies which are the permitted data formats which the event display can handle, and use this to configure the IO dialog AND what is handled via the URL?

9inpachi:
Yes, that makes sense. I will try implementing it and let you how it goes.

PR #164, opened 2020-10-29
Add more cuts

EdwardMoyse:
I noticed that the pT cuts weren't appearing for Tracks and decided to investigate
I also decided to add a few missing cuts

There are a few issues here. One, is that the cuts are very atlas-specific, so they should really be added by the JiveXML loader.

Another is that I don't think there is a way at the moment to add e.g. x,y,z cuts when the property is just a 3 dim array.


9inpachi:
All loaders extend the `PhoenixLoader`, so adding a couple of methods for defining cuts should allow us to make them experiment specific through a custom loader (like `CMSLoader`).

PR #166, opened 2020-11-02
Angular library for Phoenix UI components

9inpachi:
Hi,

## Description

This is a really big and major refactoring which moves the app components to a sub angular library for reuse. I think a review on this is very important to check if the refactoring is optimal or not.

Currently, I created the Angular library directly inside the `phoenix-app`. We also have the option of separating both the app and `phoenix-ui`, that is to have an Angular workspace at `packages/phoenix-app/` with both the app and `phoenix-ui` in separate folders inside `packages/phoenix-app/projects/`.

I opened this PR before setting up lerna for publishing the `phoenix-ui` package (I am still looking for a way to do it with the new directory structure). I want to get a review first.

## Added

* A new `phoenix-ui` package with reusable app components
* Fixed tests and refactored code according to the new package


9inpachi:
Everything is complete except the release flow with lerna. The problem is that Angular builds the actual package to be released in an entirely different folder as a different package. But we want to increment the version of the actual library as well AND THEN release the built package which is possible by using a `--contents ./dist` but then it becomes inconsistent with our `phoenix-event-display`. I will look for a better way to do it so there isn't anything unnecessary.

PR #169, opened 2020-11-11
Support to start or deploy Phoenix with a specific event

### LINKED ISSUE #149: Add functionality so that Phoenix can be started up with a specific event
The ATLAS usecase is this - physics analysis might want to have a dedicated webpage which shows 'their' event, so they can use it as part of their analysis presentation. Ideally they can also configure the setup of Phoenix to show just the objects they care about (the recent PR #148 should really help with this!)

I'm thinking that the easiest/quickest might be to have a variant of `npm run release` which takes as an argument a JiveXML file and a configuration file, and then creates an instance with this file embedded in the html. And then this can just be copied to a server.

Ideally (but this much less important) it would only contain the atlas section (this should be configurable, so other experiments can use it too)

But there may well be a better solution!

### LINKED ISSUE #163: App Deployment: make it possible to modify routes on deployment
The idea would be to be able to modify Phoenix on deployment so that only one component is shown
i.e. 
change a URL like:
http://phoenixatlas.web.cern.ch/PhoenixATLAS/#/atlas

to:
http://phoenixatlas.web.cern.ch/PhoenixATLAS/#/

i.e. the route would be modified to be (in this case):
```
const routes: Routes = [
  { path: '', component: AtlasComponent }
]
```

9inpachi:
Closes #149 
Closes #163

Hi,

## Description

This adds support for starting or deploying Phoenix with a specific event that a user can specify in a configuration file.

## Added

* A new `environment.single.ts` file for specifying configuration to deploy Phoenix with a specific event
* A configuration file `event-config.json` (`phoenix-app/event-config.json`) which specifies the event file and type to be used for Phoenix with a single event
* Scripts for starting and deploying Phoenix with a specific event
* Code for different behavior when Phoenix with a specific event is to run or deploy

## Usage

Instructions specified in [phoenix-ng README](https://github.com/9inpachi/phoenix/tree/feat-deploy-with-event/packages/phoenix-ng#readme).

For only running the app with a specific event, use `npm run start:single` from the `packages/phoenix-ng` directory.

### DIFF
--- a/packages/phoenix-ng/README.md
+++ b/packages/phoenix-ng/README.md
@@ -1,8 +1,5 @@
 # Phoenix application
 
-[![Version](https://img.shields.io/npm/v/phoenix-app.svg)](https://www.npmjs.com/package/phoenix-app)
-[![Downloads](https://img.shields.io/npm/dt/phoenix-app.svg)](https://www.npmjs.com/package/phoenix-app)
-
 This project was generated with [Angular CLI](https://github.com/angular/angular-cli) version 7.3.4.
 
 ## Build and run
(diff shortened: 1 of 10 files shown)


9inpachi:
Closes #149 and #163

PR #175, opened 2020-11-18
Fix geometry demo

EdwardMoyse:
Get the geometry demo working properly again.

The problem was the code to draw parametrised geometry was lost in the move to angular. So I re-implemented this, and added two new functions:
- buildGeometryFromParameters to EventDisplay
- addGeometryFromParameters to ThreeManager

Thanks to @9inpachi for the help!

### DIFF
--- a/packages/phoenix-event-display/src/three/index.ts
+++ b/packages/phoenix-event-display/src/three/index.ts
@@ -584,6 +584,9 @@ export class ThreeManager {
     return this.getSceneManager().getScene().getObjectByName(objectName);
   }
 
+  /** Add parametrised geometry to the scene.
+   * @param parameters The name, dimensions, and radial values for this cylindrical volume.
+   */
   public addGeometryFromParameters(parameters: any): void {
     let scene = this.getSceneManager().getScene();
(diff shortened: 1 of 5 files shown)


packages/phoenix-event-display/src/three/index.ts
```diff
@@ -579,4 +583,52 @@ export class ThreeManager {
   public getObjectByName(objectName: string): Object3D {
     return this.getSceneManager().getScene().getObjectByName(objectName);
   }
+
+  public addGeometryFromParameters(parameters: any): void {
```
9inpachi, line 587: Can you quickly add docs to this function? Or may be do it in another PR or commit.
We should have the documentation to 100% for the event display. :)

9inpachi (changes requested):

packages/phoenix-event-display/src/three/index.ts
EdwardMoyse: Good point! Thanks! 

I actually want to change `any` to specify the required parameters... but the main thing was to get this working before the presentation next week.

PR #186, opened 2020-12-04
build(event-display): update to threejs 0.122

EdwardMoyse:
Tested locally - seems fine.

PR #188, opened 2020-12-04
Add atlas default config

EdwardMoyse:
This tries to make the default for ATLAS a bit more useful.

I disabled some jet collections (to stop them overlapping and looking weird) and also some duplicate track collections. I recoloured the Pixel and SCT hits to make them standout a bit more.

As discussed in bug#187 (thanks @9inpachi!) I had to make ngOnInit async and await geometry loading.


packages/phoenix-ng/projects/phoenix-ui-components/package.json
```diff
@@ -13,7 +13,8 @@
     "dist"
   ],
   "scripts": {
-    "prepublishOnly": "ng build phoenix-ui-components --prod"
+    "prepublishOnly": "node --eval \"console.error('ERROR: Trying to publish a package that has been compiled by NGCC. This is not allowed.\\nPlease delete and rebuild the package, without compiling with NGCC, before attempting to publish.\\nNote that NGCC may have been run by importing this package into another project that is being built with Ivy enabled.\\n')\" && exit 1",
```
9inpachi, line 16: All cool apart from this. :)

9inpachi (changes requested):

EdwardMoyse:
I think this works now.

packages/phoenix-ng/projects/phoenix-ui-components/package.json
9inpachi: I meant that you need to discard this file completely (like in the previous PR #186) and not have these changes. For making use of lerna, we have to use this `package.json` for publishing but NGCC blocks that so this file should not be changed at all.

9inpachi (approved):

9inpachi:
Merging this now and will push with the changes because I think the CI is broken as well.

PR #198, opened 2020-12-17
Add more realistic geometry for ATLAS

EdwardMoyse:
This geometry was taken from a version of Tracer (https://tracer.web.cern.ch/tracer/) from June 2020 (it was not a direct copy, since back then the geometry was written in json format. Instead we converted this to gLTF).

This was initially released under the Apache 2.0 license and it is on this basis (and with the explicit permission of the ATLAS experiment) that we are including it.

This also contains some fixes for the hierarchical menus, and a new default configuration.

image


packages/phoenix-event-display/src/ui/phoenix-menu/phoenix-menu-node.ts
```diff
@@ -214,7 +214,10 @@ export class PhoenixMenuNode {
       return this;
     } else {
       for (const child of this.children) {
-        return child.findInTree(name);
+        const nodeFound = child.findInTree(name);
+        if (nodeFound) {
+          return child.findInTree(name);
```
9inpachi, line 219: Nice. I almost forgot about this.

So here, we are going through the tree again even when we have the node in `nodeFound` variable.

So just replace this line `return child.findInTree(name);` with `return nodeFound;`.

9inpachi (changes requested):

9inpachi (approved):

PR #209, opened 2021-01-31
Ability to load labels

### LINKED ISSUE #172: Add ability to load a "label" file
This is going to take a bit of thinking about ... but here are a few usecases to illustrate the point:
1. we want to be able to "decorate" Calorimeter cells with calibration information (and then possibly colour them differently, depending on the information)
2. we want to be able to identify two tracks as a positive and negative muon, as determined by the reconstruction and then add a label to the tracks in the event display
3. we want to be able to label some tracks and jets, as e.g "leading", "subleading"  and show this in the display to be able to match to a description of the event. (See attached picture for an example of this one)

It seems to me that we could easily add these to the JSON (perhaps with a new "label" key), however this will be harder for non-Phoenix formats such as JiveXML, root, or iSpy.

So instead I was thinking instead we could make it general... i.e. you can load a separate JSON file which is a bunch of labels, and then optionally display them in the display.

Perhaps the first stage could be to think of a way to add labels to objects from within Phoenix, and then save these label files? This way we would solve the format, and the UI.

For the format, I was thinking that it would need to `"Label Text" : "Type/Container Name/Position in Container"`

image

9inpachi:
Closes #172 

## Description

This adds ability to add labels to event object, save them as JSON and later load them.

## Added

* Add, remove or update labels of objects from the collections info panel
* Helper functions to easily save and load files
* Labels object kept in `PhoenixLoader` to identify object labels and load them dynamically from the Phoenix menu (JSON)
* Ability to change scale and color of labels
* Use `Object3D.getObjectByProperty` instead of traversing all children
* Labels configuration options in Phoenix menu
* Labels configuration options in dat.GUI menu
* Function to get object's position from uuid

## Screenshot

image


9inpachi:
Hi @EdwardMoyse,

This is almost ready, may be you want to check it out.
http://phoenix-pr-209.surge.sh/

EdwardMoyse:
Hi @9inpachi - thanks! Having a look now. The surge.sh link is cool - is this automatic for all CI now?

I **really** like this label implementation - I have some comments ... but these could be in future MRs?

- Ideally we would be able to change size in addition to colour etc. And maybe position along the track or something? Initially I couldn't see the label because it was very close to the centre. 
- Maybe the mouseover could say "Labels. These are added in the colleciton view"
- It would be good to link the Collection View and the Menu, e.g. by showing the label value in the text box there, and adding the Gear icon there too (so you can set colour, text size, font?) as you add it ... maybe replacing the '+' icon, because we could just make it that if the label isn't empty, it's added.

As I said though, this is GREAT!

9inpachi:
The surge stuff is all CI ;)
Although as you might have noticed it's running on my account and works
only for PRs from branches in the main repo.

I'll see what to do about the stuff you mentioned. Out of town today.

9inpachi:
**TODO:** Implement the following points as Ed suggested.

* Add tooltip to "Labels" in Phoenix menu
* Add option to configure label (color and size) directly from collections view (add a gear icon next to the label input)

PR #233, opened 2021-03-14
Allow resize, improve tests and drop improving CPU usage

9inpachi:
I am able to make it work for camera changes (`OrbitControls`) but doing it for every action is a bit more work. The render has to manually run for every action we make. That includes actions from the UI menu, Phoenix menu and animations. And it will somewhat increase complexity as well. I am not sure if it's worth it all.


9inpachi:
So we can reduce CPU usage by stopping the render when the interaction with `OrbitControls` ends (basically mouse drag stopped) and then restart the render again when the user starts the `OrbitControls` interaction (move the camera with mouse drag).

But this is only a single case and once the scene is stopped there can be other actions as well so we have to re-render every time for those actions.

The other cases of re-render will be:

1. Animation - where we have to keep rendering until animation ends
2. Phoenix menu controls
    * Color change
    * Cuts applied
    * Opacity changes
    * And much more.
3. UI menu actions
    * Clipping
    * Preset views
    * Zoom in/out
    * And much more
4. Other actions like importing an event, applying config etc.

**We have to MANUALLY re-render the scene for each of these actions I mentioned**. Which obviously is pretty hard not just to implement but to maintain as well. This will also give rise to a lot of bugs - Phoenix is big so it's hard to cover every part of code when you have to consider manual rendering.

So I say we drop this to keep everything simple, stable and maintainable.

WDYT @EdwardMoyse ?

EdwardMoyse:
> So we can reduce CPU usage by stopping the render when the interaction with `OrbitControls` ends (basically mouse drag stopped) and then restart the render again when the user starts the `OrbitControls` interaction (move the camera with mouse drag).
> 
> But this is only a single case and once the scene is stopped there can be other actions as well so we have to re-render every time for those actions.
> 
> The other cases of re-render will be:
> 
> 1. Animation - where we have to keep rendering until animation ends
> 2. Phoenix menu controls
>    
>    * Color change
>    * Cuts applied
>    * Opacity changes
>    * And much more.
> 3. UI menu actions
>    
>    * Clipping
>    * Preset views
>    * Zoom in/out
>    * And much more
> 4. Other actions like importing an event, applying config etc.
> 
> **We have to MANUALLY re-render the scene for each of these actions I mentioned**. Which obviously is pretty hard not just to implement but to maintain as well. This will also give rise to a lot of bugs - Phoenix is big so it's hard to cover every part of code when you have to consider manual rendering.
> 
> So I say we drop this to keep everything simple, stable and maintainable.
> 
> WDYT @EdwardMoyse ?

Hmm. That’s a shame - it really is noticeable how much harder my laptop has to work with Phoenix open, but I see your point. Maybe we can come back to this in the future?

9inpachi:
I am not sure, it might be harder to do in the future.

May be, for now, we can disable the loop when the browser tab is not open?

I'll take a look at the possibilities again and write down a bit of my
already existing findings on the issue just in case.

9inpachi:
Looking at the code again and trying a couple of ways, I would vote to drop this for every action - it will overcomplicate things. Just implement it for browser tab visibility.

PR #239, opened 2021-03-20
Open compressed event data

### LINKED ISSUE #232: Open compressed JiveXML
We should add the ability to open compressed JiveXML (and probably just generally add this functionality, to support e.g. compressed json or any other text format).

I started prototyping it here:
https://github.com/EdwardMoyse/phoenix/tree/support-compressed-jivexml

It doesn't work - it fails with and error about the zip file being corrupted, so I've done something wrong.

I attached theexample compressed JiveXML I was using:
[test_events.zip](https://github.com/HSF/phoenix/files/6128034/test_events.zip)

9inpachi:
Closes #232 

## Description

This adds support for importing zip files with event data in either JiveXML or JSON format.

## Added

* Ability to import event data from zip file with both JiveXML or JSON inside the zip  
  * All are loaded and available through the event selection list in UI menu
* Add and update tests for `IOOptionsDialog`
* Fix error caused when there are no track positions in JSON event data

### DIFF
--- a/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
+++ b/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
@@ -34,6 +34,9 @@ export class PhoenixObjects {
    */
   public static getTrack(trackParams: any): Object3D {
     let positions = trackParams.pos;
+    if (!positions) {
+      return;
+    }
     // Track with no points
     // if (positions.length==0) {
     //   console.log("Track with no positions.")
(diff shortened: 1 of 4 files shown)


packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
```diff
@@ -34,6 +34,9 @@ export class PhoenixObjects {
    */
   public static getTrack(trackParams: any): Object3D {
     let positions = trackParams.pos;
+    if (!positions) {
```
EdwardMoyse, line 37: The files had tracks with no positions? That's not great.

EdwardMoyse:
Looks good!

packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
9inpachi: It's this one: https://github.com/HSF/phoenix/blob/master/packages/phoenix-ng/projects/phoenix-app/src/assets/files/event_data/EventData.json

9inpachi: image

We already have RungeKutta so may be we should handle this differently.

PR #243, opened 2021-03-23
Configurable import options for event data

9inpachi:


EdwardMoyse:
Hiya. Thanks for this! It’ll be a big improvement! However I’m reading on my phone  so I’m probably missing something obvious, but I don’t quite get how you actually configure what appears... how do we set CMS io dialogues to not show Jivexml, for instance?

9inpachi:
It's not fully intuitive but we specify the formats supported by an
experiment like `<app-ui-menu [eventDataFormats]="['IG',
'JSON']"></app-ui-menu>` .

I couldn't think of another way that didn't duplicating code.

EdwardMoyse:
I think that’s fine! Is this ready to merge then?

9inpachi:
I will recheck everything tonight and merge it then.
Will push a new release afterwards.

PR #251, opened 2021-03-31
Fix calocluster position

EdwardMoyse:
This should address https://github.com/HSF/phoenix/discussions/225

N.B. I'm not totally happy with this - it would be much better to be able to use native threejs methods for all of this, so I will revisit it.


9inpachi:
For fixing the "Check Prettier" in CI.
You can do:
```
yarn prettier:write
```

PR #256, opened 2021-04-06
Add support for more types of JiveXML hits, and a new type of hits

EdwardMoyse:
This adds support for CSC, TGC, MDT, RPC, TRT hits in JiveXML

I also added a new `LineHit` and `getLines` functionality (not very happy with these names)

Currently these go under a new menu item `Line Hits` but really I think it would be better if they went under `Lines`.

I updated the default ATLAS configuration.

Picture showing TRT hits:
image


packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
```diff
@@ -277,11 +324,16 @@ export class PhoenixObjects {
...
-    // if (cube.position.x * cube.position.x + cube.position.y * cube.position.y > maxR * maxR) {
-    //   cube.position.x = maxR * Math.cos(clusterParams.phi);
-    //   cube.position.y = maxR * Math.sin(clusterParams.phi);
-    // }
-    // cube.position.z = Math.max(Math.min(pos.z, maxZ), -maxZ); // keep in maxZ range.
+    // FIXME - more elegant way to do this? Maybe natively use cylindrical here?
+    // How to generalise? Pass in limit lambda?
+    if (
+      cube.position.x * cube.position.x + cube.position.y * cube.position.y >
+      maxR * maxR
```
9inpachi, line 331: Not a more elegant way but using JavaScript's built-in methods, may be we can shorten it.

```ts
Math.pow(cube.position.x, 2) + Math.pow(cube.position.y, 2) > Math.pow(maxR, 2)
```

9inpachi (approved):
The code looks really good to me.

One thing I noticed that the "Line Hits" are enabled by default but no children are enabled.

image

packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
EdwardMoyse: By “elegant” I meant that this is currently hard coded for ATLAS and we should generalise. Not quite sure how to do it. I can change to use Math.pow but for simple squares I thought this was faster?

9inpachi:
Oh I see.

No need fo change it. I am not sure about rhe performance myself.

PR #257, opened 2021-04-06
Options for CaloClusters

### LINKED ISSUE #253: Add scaling for CaloClusters
It would be great to add a scaling for CaloClusters, so we can adjust how large they are. The amount of energy deposited on average can depend on the specific run - so this is likely to need tweaking. 

I would propose adding a "collection level" settings gearbox for CaloClusters. It might be good to add options like opacity and wireframe etc.

image

9inpachi:
Closes #253 

Added scaling for CaloClusters and opacity and wireframing options for all collections.

### DIFF
--- a/packages/phoenix-event-display/src/loaders/phoenix-loader.ts
+++ b/packages/phoenix-event-display/src/loaders/phoenix-loader.ts
@@ -165,13 +165,14 @@ export class PhoenixLoader implements EventDataLoader {
         typeFolder: any,
         typeFolderPM: PhoenixMenuNode
       ) => {
+        const scaleJets = (value: number) => {
+          this.graphicsLibrary.getSceneManager().scaleJets(value);
+        };
         if (typeFolder) {
           const sizeMenu = typeFolder
             .add({ jetsScale: 100 }, 'jetsScale', 1, 200)
             .name('Jets Size (%)');
-          sizeMenu.onChange((value: number) => {
-            this.graphicsLibrary.getSceneManager().scaleJets(value);
-          });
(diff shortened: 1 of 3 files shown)


9inpachi:
@EdwardMoyse Let me know if this isn't fine, we can reopen the issue.

EdwardMoyse:
Sorry for not replying earlier! This is great, but one improvement would be
to just scale the radial size. Normally the eta/phi represent the size of
the Calo cell (which is fixed( and the radial is the energy (where the
relative importance depends on the physics).

(It’s so great that I can just check this in my phone btw - this deployment
magic you did in the CI is really fantastic!)

9inpachi:
Sorry but I don't quite get what it means by radial size, do you mean the
length of the calo cell from the direction of origin? Like the length that
currently is decided by energy?

BTW, I was thinking we should have a "dev" version of Phoenix deployed
somewhere as well. Ever since we moved to releases, the production version
isn't always updated. We could have a "dev" version which we update with
each commit to master. WDYT?

EdwardMoyse:
On Sat, 10 Apr 2021 at 01:34, Fawad Ali ***@***.***> wrote:

> Sorry but I don't quite get what it means by radial size, do you mean the
> length of the calo cell from the direction of origin? Like the length that
> currently is decided by energy?


Yes, exactly that.

>
>
> BTW, I was thinking we should have a "dev" version of Phoenix deployed
> somewhere as well. Ever since we moved to releases, the production version
> isn't always updated. We could have a "dev" version which we update with
> each commit to master. WDYT?


Good idea!



>

PR #260, opened 2021-04-12
Extend tracks using hits, and hit code cleanup

EdwardMoyse:
Various improvements/cleanup related to drawing hits and tracks, specifically, try to use hits to extend short tracks.

Honestly, this doesn't work as well as I'd hoped - whilst the muon tracks now reach into the muon system, they look very "kinked" ... this is, I think, due to the positions in the JiveXML not being the best ...

I also cleaned up the hit drawing code, and added some extra user-data from the JiveXML

Future work:
- make it possible to turn this extension off/on 
- move some of the code to link hits/tracks to phoenix proper and generalise (then it can be used to e.g. colour hits by the track they're associated to)
- render hits as clusters and strips
- add the ability to draw muon chambers (this will likely need to be from the JSON)

jivexml-loader:
- Store hits associated with tracks (would be nice not to show this in the collections view)
- Add new function `jivexml-loader:getPositionOfHit` using hit identifiers
- Rework how we store hits, so eventually we can maybe cut on them
- Remove calls to `HitLines`, since now Line hits are just another type of hit

phoenix-loader:
- Rework how we handle hits to remove `HitLines`. Added ability to `concatonateObjs` in `addObjectType` and `addObjectCollection`, so we can still benefit from the performance of having one `BufferGeometry` per Hit type.

phoenix-objects:
- `getHits` can now handle types (default is 'Point')
- added new `hitsToPoints` and `hitsToLines` to handle the two current cases


EdwardMoyse:
@9inpachi I'm really not sure why this is failing. I guess I need to improve the documentation? I can't run locally however. I have `compodoc` installed, but `yarn run docs:coverage` fails with:
```
$ lerna run docs:coverage --scope phoenix-event-display
lerna notice cli v3.22.1
lerna notice filter including "phoenix-event-display"
lerna info filter [ 'phoenix-event-display' ]
lerna info Executing command in 1 package: "yarn run docs:coverage"
phoenix-event-display: $ yarn compodoc -p configs/compodoc.conf.json --coverageTest 100
phoenix-event-display: error Command "compodoc" not found.
phoenix-event-display: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
phoenix-event-display: error Command failed with exit code 1.
phoenix-event-display: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
lerna ERR! yarn run docs:coverage exited 1 in 'phoenix-event-display'
lerna ERR! yarn run docs:coverage stdout:
$ yarn compodoc -p configs/compodoc.conf.json --coverageTest 100
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

lerna ERR! yarn run docs:coverage stderr:
error Command "compodoc" not found.
```

packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
```diff
@@ -215,33 +202,57 @@ export class PhoenixObjects {
...
-    let hitsParamsClone: any;
-
-    // If the parameters is an object then take out 'pos' for hits positions
-    if (typeof hitsParams === 'object' && !Array.isArray(hitsParams)) {
-      positions = [hitsParams.pos];
-      hitsParamsClone = hitsParams;
-    } else {
-      positions = hitsParams;
-      hitsParamsClone = { pos: hitsParams };
+  public static getHits(hitsParams: [{ pos: []; type?: string }]): Object3D {
```
9inpachi, line 209: This `hitsParams` is a single parameter and `type` is inside a single object in array. I am guessing that this doesn't parse `@returns` correctly because of an extra `@param` and throws the docs error.

9inpachi (changes requested):

9inpachi:
>  I can't run locally however. I have compodoc installed, but yarn run docs:coverage fails with:

Did you do `yarn install:dependencies`? There shouldn't be a need to install compodoc separately and the CI also just picks up compodoc after just `yarn install:dependencies` in `phoenix-event-display`.

packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
```diff
@@ -215,33 +202,57 @@ export class PhoenixObjects {
...
+          hitsParams,
+          hitsParamsClone
+        );
+      default:
+        console.log('ERROR: Unknown hit type!');
+        return;
+    }
+  }
+
+  private static hitsToPoints(
```
9inpachi, line 251: Missing docs.

packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
```diff
@@ -252,57 +263,34 @@ export class PhoenixObjects {
...
-    const pointPos = new Float32Array(positions.length * 3);
-    let i = 0;
-    for (const hit of positions) {
-      pointPos[i] = hit[0];
-      pointPos[i + 1] = hit[1];
-      pointPos[i + 2] = hit[2];
-      i += 3;
-    }
-
+  private static hitsToLines(
```
9inpachi, line 274: Missing docs.

packages/phoenix-event-display/src/loaders/jivexml-loader.ts
```diff
@@ -447,79 +414,51 @@ export class JiveXMLLoader extends PhoenixLoader {
   public getMuonPRD(
```
9inpachi, line 414: Missing docs for `name` param.

9inpachi (changes requested):

9inpachi (approved):

EdwardMoyse:
Thanks a lot! Got caught up in mealtime....

PR #271, opened 2021-04-14
feat(event-display): add ability to scale clusters in just one direction

EdwardMoyse:
Follow up to discussion on PR #257

Allows to scale clusters just in z direction (but made this a general feature of `scaleChildObjects`)


9inpachi:
Great. I should have done this from the start. (Even though I knew it - I think I forgot to)

Issue #266, opened 2021-04-14
Add some guidelines


EdwardMoyse:
We already have 'show axis' under View controls e.g.
https://github.com/HSF/phoenix/blob/master/packages/phoenix-ng/projects/phoenix-ui-components/src/components/ui-menu/view-options/view-options.component.ts

However it would be useful to be able to show a labelled eta/phi or theta/phi grid. Ideally this would be adjustable in the GUI.

See here for the VP1 example:
http://atlas-vp1.web.cern.ch/atlas-vp1/doc/doc_guidelines/

(though we could probably do better)

9inpachi:
@EdwardMoyse Can this be closed?

PR #273, opened 2021-04-15
feat(event-display): add guidelines and simplify some coordinate tran…

EdwardMoyse:
For the simplifying of the coordinate transformation, as discussed in the last Phoenix weekly, this is not the end of this story .. we should make these transformation configurable for other experiments. But it's definitely better (IMO)

Also, tweaked some default cuts.

Finally, this adds a first go at guidelines as discussed in #266

image

### DIFF
--- a/packages/phoenix-event-display/src/helpers/coordinate-helper.ts
+++ b/packages/phoenix-event-display/src/helpers/coordinate-helper.ts
@@ -1,4 +1,4 @@
-import { Vector3 } from 'three';
+import { Vector3, Quaternion } from 'three';
 
 /**
  * Helper methods for coordinate conversions.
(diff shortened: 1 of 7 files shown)


9inpachi:
The pipeline is broke because of the added functions not documented.

I think it's important to have an explanation for all the functions. The name themselves don't explain that well. And since we are aiming to be a framework, it's a lot easier to work with one if it has docs available on hover in the IDE - which we do get. ;)

WDYT? We could always remove this check if it's a bother.

EdwardMoyse:
I think it’s a good idea - I just wish compodoc listed the functions which
failed. I worked it out by trial and error locally.

EdwardMoyse:
Hmm. I didn’t mean to merge this with CI failures! I’m not sure what happened.

PR #275, opened 2021-04-16
feat(event-display): improve grid lines

EdwardMoyse:
Add better labels
Colour the lines to match the axes
Dashed lines to make them clearer

image


EdwardMoyse:
Wow. Thanks @9inpachi  - can you explain why these imports were necessary?

9inpachi:
They weren't necessary but apparently made their way in through the IDEs suggestions.

EdwardMoyse:
Can we revert the IDE changes? Which were they? I just want to understand so I don’t do it again ...

9inpachi:
Seems like this was the commit https://github.com/HSF/phoenix/commit/56f94f043b5bdf5e521c05e0bc058186c78270ad#diff-90c13675ebee5685b8bbe8893b2403e971f854d3521d2778c7f5472d915c5335

We can't really automatically revert them since it's sort of a human error. The normal scenario is that you were normally typing and pressed enter on one of the IDE suggestions and then removed it with backspace since that wasn't what you wanted. Even if you did remove it manually, the import didn't go away.

One thing I do to make sure is that when I use some IDE suggestion and realize that it's not the one I want, I use `Ctrl + z` to undo that change and that also removes the import since it was done simultaneously with the enter.

PR #280, opened 2021-04-18
Color options for event data objects

9inpachi:
## Description

This adds options for colors based on momentum and charge for tracks. The color options are generic so they can be extended to use for other event data as well.

## Changes

* Create `ColoringManager` for color operations using three.js
* Create `ColorOptions` class for managing color options in the UI
  * Add functionality to color by momentum (value from 1 to 50000)
    The min and max are separated from the median for color
  * Add functionality to color by charge (-1, 0, 1)
* Add track parameters to `Line` object to be able to have colors on zoom out
* Add types for dat.GUI
* Add theming CSS for `select`
* Add `select` config type for Phoenix menu item


EdwardMoyse:
Hi @9inpachi - that looks like a really good start. I didn't manage to get the colour by momentum working though

9inpachi:
Yeah, I haven't had the chance to progress there. I just wanted to confirm if the UI was right for momentum colors with that min and max range. And also need to know the unit of momentum I mentioned in the original issue.

EdwardMoyse:
Sorry, for units atlas (and CMS) typically have tracks in the 0.3 GeV to 1 TeV range. So I think 1-1000 GeV would be a good start. We might need to configure this for other experiments later.

9inpachi:
@EdwardMoyse I have been trying to make this work but a "clean" solution is kind of hard. For now, everything works as intended, so may be you want to check out if this is expected. I will create a class to handle everything (later - I am drained for the day :D) and that might make things cleaner and generic (hopefully).

I am just thinking the UI doesn't look too appealing for all the options we are adding. May be it is better to do it in a dialog rather than the Phoenix menu. That will require significantly more effort and we will lose the perks of Phoenix menu but I suppose the UI will be simpler and easier. WDYT?

EdwardMoyse:
Hi @9inpachi - I think we can probably manage without a dialog if you make this more compact. For instance, could you hide the options not related to the current 'colour by' option? For example, if you select colour by momentum, then the momentum options appear. If you change to colour by charge, then the boxes for charge colour appear.

9inpachi:
@EdwardMoyse I moved the color options inside a child node. May be I haven't pushed that yet - I will. That makes it really compact now and easy to use.

EdwardMoyse:
Okay, that looks good - we could then move the cuts into a node too? Because it won't make sense to have the base colour in the gearbox, along with cuts, and the rest of the colours there. So we could maybe have : colour options, draw options (wireframe, opacity etc) and cuts?

9inpachi:
That sounds good. I was thinking somewhat in the same direction.

You might have noticed Phoenix getting slower if there are too many options open in Phoenix menu. I think we can reduce that by dividing options in nodes because the nodes render the configuration options so if the nodes aren't loaded/visible, the configuration options don't render saving us some performance.

And I think I will rename all the `color` instances to `colour`. Since that's the standard for Europe. (I studied `colour` myself but I guess US English is having its influence. :D)

9inpachi:
Keeping the spelling `Color` to keep consistency because that's what three.js uses.

EdwardMoyse:
And because it’s the correct spelling ;-)

PR #298, opened 2021-05-03
Color by Vertex option

9inpachi:
Hi,

## Description

This adds color by vertex option and types for Phoenix menu. It also only shows configs for selected color by option.

## Changes

* Add detailed types for Phoenix menu configurations
* Add option to color tracks by Vertex
* Add ability to filter and show only a selected group of configs
* Fix tests not working for fullscreen

### DIFF
--- a/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
+++ b/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
@@ -20,7 +20,6 @@ import {
   SphereBufferGeometry,
   LineSegments,
   LineDashedMaterial,
-  Spherical,
 } from 'three';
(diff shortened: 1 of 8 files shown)


9inpachi:
@EdwardMoyse — the current implementation uses a random color for each vertex to color the tracks that are associated with it. I don't suppose the option to select the color for each vertex manually is viable? We could do it through the collections info panel but the conditional rendering needed to show the option only for vertices will surely be slowing things down a bit.

EdwardMoyse:
I think random colour per vertex is fine, at least for now. The only thing might be that there is normally a primary vertex, so maybe that should have a specific colour.

9inpachi:
@EdwardMoyse This took a lot of banging my head against the wall (both Phoenix menu types and showing only selected config options) but it finally worked. :D

You can take a look if you want.

PR #322, opened 2021-06-10
Misc atlas fixes

### LINKED ISSUE #318: Jets aren't proportional to their energy.
Like Atlantis, the jets don't have a proportional size to their energy. Both Atlantis and Phoenix jets are shown in the attached images (same xml file).
In Phoenix, id:0 and id:1 are the two leading jets.
phoenix_jet
Atlantis_jet

### LINKED ISSUE #319: Phoenix does not display photon and electron objects
Phoenix menu does not display photon and electron objects using .xml input.
On the Event data collection info tab, these object information may be viewed but they cannot be highlighted.

EdwardMoyse:
Includes many fixes such as:
- no longer limit jet lengths since we can scale them bb41683 (which fixes #318 )
- better handling of JiveXML objects (pt and fix cuts to match, out-of-bounds angles etc)
- handle compound objects consistently (and added electron and photon which fixes #319 )
- refactored some elements of the runge kutta

### DIFF
--- a/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
+++ b/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
@@ -44,21 +44,14 @@ export class PhoenixObjects {
     // Track with too few points are extrapolated with RungeKutta
     if (positions.length < 2) {
       if (trackParams?.dparams) {
-        // Test, for ATLAS.
-        // FIXME - make configurable
-        let inBounds = function (pos: Vector3) {
-          if (pos.z > 3000) return false;
-          if (Math.sqrt(pos.x * pos.x + pos.y * pos.y) > 1100) return false;
-
-          return true;
-        };
-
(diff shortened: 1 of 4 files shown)


packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
```diff
@@ -150,10 +140,6 @@ export class PhoenixObjects {
       : CoordinateHelper.etaToTheta(eta);
     // Jet energy parameter can either be 'energy' or 'et'
     let length = (jetParams.energy ? jetParams.energy : jetParams.et) * 0.2;
-    // Ugh - We don't want the Jets to go out of the event display
```
9inpachi, line 153: Is this fine for now? I was thinking of normalizing the energy in a certain range. The differences should then be apparent and Jets won't scale more than the event display.

9inpachi (approved):

PR #332, opened 2021-07-08
Use DOM overlay for AR

### LINKED ISSUE #331: Add controls in AR
This might be the solution:

https://discourse.threejs.org/t/how-to-add-a-ui-to-immersive-xr-ar-view/16532/6

9inpachi:
Closes #331 

## Description

This makes overlays like Phoenix menu and UI menu available in AR mode.

## Changes

* Update docs with latest features and add AR to docs
* Add flag to enable DOM overlay in AR mode (`ARManager.enableDomOverlay`)
* Add menu to AR toggle to enter AR with or without overlays


EdwardMoyse (approved):
Looks good to me!

9inpachi:
@EdwardMoyse I haven't tried this out yet. Unfortunately I don't have a device with AR support at hand and this doesn't work with the WebXR extension. Since it's still just a draft in W3C so I am not sure if it will even work.

EdwardMoyse:
Yeah it's a shame it wasn't deployed so I can't easily test it. Later I could try to run locally and test, or we could accept it and cross our fingers.

9inpachi:
No use merging either, because we won't have it deployed to the main hepsoftwarefoundation site that has https.

EdwardMoyse:
> 
> 
> No use merging either, because we won't have it deployed to the main hepsoftwarefoundation site that has https.

Well, actually I tried AR with http:// and it seemed to work okay.

9inpachi:
Really? That seems weird. Then I can just deploy this PR and you can try it out.

EdwardMoyse:
Hmm. It sort of works - a few glitches, like the collision animation just hides the event data (I guess other animation modes should be hidden in XR). But the main problem is the AR is disabled - just a white background. It’s still very cool! But no longer AR.

9inpachi:
> Hmm. It sort of works - a few glitches, like the collision animation just hides the event data (I guess other animation modes should be hidden in XR). But the main problem is the AR is disabled - just a white background. It’s still very cool! But no longer AR.

I see. Good to know, the white background seems to be the main DOM background. We can make it transparent no problem. I'll keep this open for when I get the device. Or maybe I should just buy a new phone. But argh, I don't need it for any other thing.

9inpachi:
@EdwardMoyse thanks for the video. I have an idea of why the background is so.

One thing I am a bit undecisive about is if we want to have another option to have AR mode without overlays? It's also possible to have it like this with the overlay and use the screenshot mode to hide everything but I think having the option to not show overlays to begin with is a good idea too. We can have the two options (with overlay and without overlay) as pop-up like in view options. WDYT?

EdwardMoyse:
I definitely like the idea of having a choice, even if it's only something we set in configuration before hand (ideally though, we get a choice when we select AR)

9inpachi:
@EdwardMoyse could you try this out now?

EdwardMoyse:
I still have the white background with overlays, as expected, but without works just as before. So this looks great! Thanks!

9inpachi:
It seems to be working for me in S7 Edge. @EdwardMoyse could you try this on your phone now? It would be better if you could test with the https environment.

EdwardMoyse:
Hi @9inpachi - I tried (with https://) and I'm afraid it's still the same ... white background with overlay.

9inpachi:
@EdwardMoyse this might be an iOS issue. Could you try out any other three.js example with overlay in AR.

EdwardMoyse:
Sorry for the late reply. I just tried the Cones AR demo and it works fine. So I don’t think it’s necessarily an iOS issue. 


https://user-images.githubusercontent.com/6764617/126064652-5fb0d9bb-3a47-4cb8-9361-97e588e6a997.MP4

9inpachi:
@EdwardMoyse interesting. They had `display; none` as style for the overlay, maybe that's the catch.

Could we pair on this sometime? I don't have access to an iOS device at all which makes the process dependent and slow.

PR #341, opened 2021-07-20
Fix AR overlay for iOS

9inpachi:


EdwardMoyse:
Hi Fawad - I just tried this and now I don't seem to have **any** overlays...

9inpachi:
Worked on Android though. Let's pair up for this. I don't see any other way of fixing this.

9inpachi:
@EdwardMoyse, I think we can add this. If you get an hour or two to pair program, let me know.

EdwardMoyse:
Hi @9inpachi I could do tomorrow afternoon?

9inpachi:
That won't work for me. Any time during work hours would be difficult.
Maybe before 10 in the morning and after 6 in the evening.

PR #355, opened 2021-09-18
Do not apply Phoenix menu node toggle if undefined

9inpachi:
With this we can manually remove parts of the Phoenix menu configuration and then the value from code will be used.

@EdwardMoyse This will be useful for managing geometries through URL options.


EdwardMoyse:
Thanks! This is great.

PR #381, opened 2021-11-27
Fix jets disappearing on scaling to 0

### LINKED ISSUE #379: Scaling Jets to zero makes them vanish and then they cannot be recovered
As above - I assume this is because once they have zero length, this cannot be recovered. If this is true, perhaps we can set a minimum length so we can always rescale them?

This is one of the reasons I want to try to think about a new way of drawing objects - i.e. when we change cuts / draw options we can completely redraw the collection. I'll open a separate ticket about this.

9inpachi:
Closes #379


9inpachi:
Not sure why the deployment status didn't update, but the PR is deployed at http://phoenix-pr-381.surge.sh.

PR #412, opened 2022-03-13
Add LHCb SciFi hits event data visualization

andrewpap22:
Hi all, 

As promised here is the pull request for the SciFi hits data visualization. 

It seems that I'm in a good way to finish it, since the data are being read and stored properly and I got to the point into visualising "something". Now, that "something" is not what I was expecting. Probably Mr. @sponce can help me a bit more on this, since I think I'm manipulating the data in a wrong way. Please do review the following part of the code in the file path: `packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts` lines: `606 - 609`

```javascript
     /// we need 6 coords to draw the line:
    /// (x0, y[0], z0, x0+dxDy*(y[1]-y[0]),y[1],z0)

    const fiberHits = [];
    fiberHits.push(new Vector3(0, 0, 0));
    fiberHits.push(new Vector3(x0, y0, z0));
    fiberHits.push(new Vector3(x0 + dxDy * (y1 - y0), y1, z0));
```

First of all the position that the "lines" a.k.a fibers are being displayed is way out of the FT detector and along that I'm not sure they are being visualized in the correct direction as well. You can see what I mean in the picture below: 

Screenshot 2022-03-13 at 3 41 16 PM

So yeah I will try to fix that, and then another concern I have is, (and on that maybe you can help me a bit more dear @9inpachi) wether the way I've written the code is the proper way to go, and by that I mean that this particular code is not generic to be used by others even though I'm adding it in the global Phoenix physics objects, it is LHCb specific I think. If there is a way to make it generic and keep it there, I'm glad to make it, but I'll need some help on that, or if you think that it should be moved somewhere else, please do ping me and I'll fix it asap. 

Thank you all!

### DIFF
--- a/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
+++ b/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
@@ -586,4 +586,22 @@ export class PhoenixObjects {
 
     return object;
   }
+
+  /**
+   * Process the Fibers (a.k.a reconstructed line from points / hits) from the given parameters and get it as a geometry.
+   * @param Fiber Parameters for the SciFi Calorimeter.
+   * @returns Fiber object.
+   */
+  public static getFibers(FiberParams: any): Object3D {
+    /// retrieve and store the data
+    const fiberHits = [];
+    fiberHits.push(new Vector3(0, 0, 0));
+    fiberHits.push(new Vector3(fiberHits.x0, fiberHits.z0, fiberHits.dxDy));
+
+    /// for my 6 coords to draw the line: 
+    /// (x0, y[0], z0, x0+dxDy*(y[1]-y[0]),y[1],z0)
+
+    /// return (FiberObject)
+    return new Object3D();
+  }
 }
--- a/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
+++ b/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
@@ -593,15 +593,39 @@ export class PhoenixObjects {
    * @returns Fiber object.
    */
   public static getFibers(FiberParams: any): Object3D {
-    /// retrieve and store the data
-    const fiberHits = [];
-    fiberHits.push(new Vector3(0, 0, 0));
-    fiberHits.push(new Vector3(fiberHits.x0, fiberHits.z0, fiberHits.dxDy));
+    /// retrieve and store the coords to draw the fibers
+    const x0 = FiberParams.x0;
+    const y0 = FiberParams.y[0];
+    const y1 = FiberParams.y[1];
+    const z0 = FiberParams.z0;
+    const dxDy = FiberParams.dxDy;
 
-    /// for my 6 coords to draw the line: 
+    /// we need 6 coords to draw the line:
     /// (x0, y[0], z0, x0+dxDy*(y[1]-y[0]),y[1],z0)
 
+    const fiberHits = [];
+    fiberHits.push(new Vector3(0, 0, 0));
+    fiberHits.push(new Vector3(x0, y0, z0));
+    fiberHits.push(new Vector3(x0 + dxDy * (y1 - y0), y1, z0));
+
+    /// geometry
+    const geometry = new BufferGeometry().setFromPoints(fiberHits);
+
+    /// material
+    const material = new LineBasicMaterial({
+      linewidth: 2,
+      color: FiberParams.color ?? EVENT_DATA_TYPE_COLORS.Fibers,
+    });
+
+    /// object
+    const fiberLines = new LineSegments(geometry, material);
+
+    fiberLines.userData = Object.assign({}, FiberParams);
+    fiberLines.name = 'FiberLine';
+    /// Setting uuid for selection from collections info
+    FiberParams.uuid = fiberLines.uuid;
+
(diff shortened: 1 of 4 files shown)


andrewpap22:
I think I just fixed the visualization problem with the following commit. 

Please Mr. @sponce do check if it looks proper to you. Right now it is inside the SciFi detector but I'm still not sure of the direction of the lines based on some Panoramix images that I've checked. After we display it 100% correctly we can try and make some dots on top of the lines to visualize the hits that the fibers are being created of. 

Have a look on the following pictures: 

Screenshot 2022-03-13 at 3 50 38 PM
Screenshot 2022-03-13 at 3 52 19 PM

packages/phoenix-event-display/src/loaders/phoenix-loader.ts
```diff
@@ -290,6 +290,16 @@ export class PhoenixLoader implements EventDataLoader {
       );
     }
 
+    if (eventData.Fibers) {
```
9inpachi, line 293: We already have an [`LHCbLoader`](https://github.com/HSF/phoenix/blob/master/packages/phoenix-event-display/src/loaders/lhcb-loader.ts) which this code should belong to.

Since we are already extending the `LHCbLoader` from `PhoenixLoader`, what you can do is put the method there, and override the [`loadObjectTypes`](https://github.com/HSF/phoenix/blob/b1a88476b5a73412f95f0360bdad4c099bed917d/packages/phoenix-event-display/src/loaders/phoenix-loader.ts#L145) method on `LHCbLoader`, call the parent version of it and then add your LHCb specific code there.

The `CMSLoader` does exactly the same [here](https://github.com/HSF/phoenix/blob/master/packages/phoenix-event-display/src/loaders/cms-loader.ts#L28). For CMS, we are adding `MuonChambers` as a custom object.

Now that I wrote this comment, I just realized that I wrote a dedicated guide for this. If you just read through this, you will have a good idea of what to do: <https://github.com/HSF/phoenix/blob/master/guides/developers/event-data-loader.md>

It will look something like this in the `LHCbLoader`:

```ts
  protected loadObjectTypes(eventData: any) {
    // Load all other object types.
    super.loadObjectTypes(eventData);

    // Load LHCb specific object types.
    if (eventData.Fibers) {
      this.addObjectType(
        eventData.Fibers,
        LHCbObjects.getFibers,
        'Fibers',
        false
      );
    }
  }
```

packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
```diff
@@ -586,4 +586,46 @@ export class PhoenixObjects {
 
     return object;
   }
+
+  /**
+   * Process the Fibers (a.k.a reconstructed line from points / hits) from the given parameters and get it as a geometry.
+   * @param Fiber Parameters for the SciFi Calorimeter.
+   * @returns Fiber object.
+   */
+  public static getFibers(FiberParams: any): Object3D {
```
9inpachi, line 595: Since you mentioned this is only specific to LHCb, we should have a dedicated class for `LHCbObjects` just like we have for [`CMSObjects`](https://github.com/HSF/phoenix/blob/master/packages/phoenix-event-display/src/loaders/objects/cms-objects.ts). Everything works pretty much the same but in the [`CMSLoader`](https://github.com/HSF/phoenix/blob/master/packages/phoenix-event-display/src/loaders/cms-loader.ts), we call `CMSObjects.getMuonChamber` instead of `PhoenixObject.getMuonChamber`.

This guide is exactly what you are looking for: https://github.com/HSF/phoenix/blob/master/guides/developers/event-data-loader.md#handling-new-physics-objects

9inpachi (changes requested):
Hi @andrewpap22,

Thanks for sharing the details.

From what I have understood, what you are doing is adding an experiment specific physics object. I have written a guide on this which should help you understand the following:

1. How to create a custom event data loader for a specific experiment (the `LHCbLoader` already exists in your case) and
2. How to add custom physics objects for a specific experiment. (`LHCbObjects` doesn't exist)

Just go through this guide: https://github.com/HSF/phoenix/blob/master/guides/developers/event-data-loader.md

EdwardMoyse:
Hi all,

Sebastian and I discussed this just now, and we think that rather than introduce a completely new type, it would make more sense to just use the existing 2D hits (i.e. lines)

For an example from ATLAS of this in use, see here:
https://github.com/HSF/phoenix/blob/b1a88476b5a73412f95f0360bdad4c099bed917d/packages/phoenix-event-display/src/loaders/jivexml-loader.ts#L617-L630

Essentially it's just like a normal hit object, but instead of 3 coordinate for a point, we use 6 coordinates for a `LineHit`.

sponce:
The LHCb code has now been modified to use 2D hits for this case, so let's close this PR

sponce:
@EdwardMoyse it seems I do not have the rights to close this. Can you close it yourself ?

EdwardMoyse:
I sent you an invite to join the project, then you should be able to close
(I can do it of course, but this would be a good check that you have been
added correctly).

sponce:
I'm not sure I understand this comment of 4 days ago. This was all sorted out in April, and I indeed closed it myself at that time.

EdwardMoyse:
This might be because we had to restore the repository?

PR #419, opened 2022-04-14
Load all scenes of a gltf file, not only the main one

sponce:
The adding of geometries to the UI menu has been moved to the callback so that it becomes possible to create several geometries from a given file, one per scene.
The name provided is now disregarded if it's empty, and the names of the scenes in the geometry are used in such a case.
They may contain a path with the ' > ' syntax, these path will be respected to build a hierarchy in the menu


packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
```diff
@@ -208,29 +208,47 @@ export class ImportManager {
...
+            var menuName = menuNodeName;
+            if (pos > -1) {
+              if (menuName != "") menuName = menuName + ' > ';
+              menuName = menuName + geoName.substring(0, pos).replace('_>_', ' > ');
+              geoName = geoName.substring(pos + 3);
+            }
+            this.processGeometry(geometry, geoName, scale);
+            var visible = initiallyVisible;
+            if ("visible" in geometry.userData) visible = geometry.userData["visible"];
+            callback(geometry, geoName, menuName, visible);
```
9inpachi, line 248: As I understand it, all the scenes will be loaded as separate geometries? The geometries are then added to the menu separately as well?

9inpachi: Could you also please tell me in what format the geometry names are set in the glTF file? The separator seems to be `_>_`.

9inpachi: I think there may be a better way to handle this instead of going through all these callbacks which makes the code hard to read and understand.

I can think of something once I get an answer to the questions above.

9inpachi:
It would be great if you could also set up linting as mentioned in our [CONTRIBUTING guidelines](https://github.com/HSF/phoenix/blob/master/CONTRIBUTING.md#code-formatting). Otherwise, the pipeline fails.

packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
sponce: Yes, all scenes will be loaded as separate geometries, very much like it's done now actually, except we allow to have them all in a single file and do not force users to have n files, one per geometry. As gltf format is having this functionality, it would be sad not to use it.

Concerning format, I've just kept what was used so far, so ' > ' separator. The thing however is that spaces in the gltf file seem to be replaced by underscores at some level, so it ends up being "_>_". No clue why.

Concerning callbacks, I support your position, but again tried to follow the style of what was there, that is using callbacks everywhere. Seems to be also a generic thing in javascript.

sponce:
For linting, will do it and update the MR. Just out of curiosity : why forcing all developers to install it locally and not doing it on the server as a ci job ?

PR #425, opened 2022-04-25
Fix opacity not being set for LHCb geometry

9inpachi:
This fixes the opacity not being set for LHCb geometry but introduces transparency problems in ATLAS.


9inpachi:
Test at http://phoenix-pr-425.surge.sh

9inpachi:
Hi @EdwardMoyse, 

I think you should test this if it looks okay.

EdwardMoyse:
Sorry for the delay @9inpachi. Can you clarify what transparency problems it introduces? I'm about to open a ticket about one I spotted, but this is in production (i.e. before the PR is merged).

9inpachi:
If you compare the Jets in ATLAS in the deployed PR and the production
version, you will immediately notice.

EdwardMoyse:
Okay, now I see it. Yeah, unfortunately I don't think we can let this in as-is. I will try to have a look myself next week.

sponce:
Any news on that ? It works nicely locally for me (and LHCb visualization) and would be nice to have it fixed in master.

packages/phoenix-event-display/src/managers/three-manager/scene-manager.ts
```diff
@@ -404,6 +404,16 @@ export class SceneManager {
   public eventDataDepthTest(value: boolean) {
     const object = this.getEventData();
 
+    // Remove transparency from detector geometry for depthTest to work
+    if (!value) {
```
9inpachi, line 408: Not need for this.

packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
```diff
@@ -353,6 +353,7 @@ export class ImportManager {
             color: color,
             shininess: 0,
             side: side,
+            transparent: true,
```
9inpachi, line 356: Pass it as an argument.

sponce:
@9inpachi your last version was almost fine, but for one little bug when calling processGeometry (missing argument) and for the handling of user defined opacity. I've fixed this in a PR on your PR. Can you review, check it works fine with Atlas and merge everything ?

9inpachi:
It seems introducing the loading of all scenes in GLTF has introduced some bugs for detector geometry of ATLAS. Will need to look into that.

sponce:
I can have a look at the problems in Atlas if you tell me how to reproduce

9inpachi:
It's not a bug in the geometry itself but how the menu and glTF scenes are linked now. Since we are loading all scenes now, the Phoenix glTF files are not structured correctly. I think a way would be to have a different function for loading single scenes.

There are all these auxiliary scenes in the menu now that don't do anything and the actual controls for the geometry parts are gone.

image

sponce:
This is weird, the behavior was supposed to be backward compatible. Let me have a look

9inpachi:
Sorry, this might be because of the refactoring I did. I will have a look
tonight.

sponce:
I confirm the problem was introduced in https://github.com/HSF/phoenix/commit/6b258880db9928a19a9d8ef0c26fdedfc78610fc. I had a line saying : 
```javascript
 let geoName = name === '' ? geometry.name : name;
```
and thus respecting the name passed when one was given (the case for Atlas). Now it has become : 
```javascript
  const sceneName = this.processGLTFSceneName(scene.name);
```
thus always taking the name from inside the geometry (called scene here) and ignoring the name given in the call.
The fix may be as simple as : 
```javascript
  const sceneName = name === '' ? this.processGLTFSceneName(scene.name) : name;
```

sponce:
Nope, this change has broken all menus of LHCb somehow ! Please revert

sponce:
actually I do not get how your function `processGLTFSceneName` works. It returns `{ name, menuNodeName }`, and then it's used that way : 
```
const sceneName = this.processGLTFSceneName(scene.name);
this.processGeometry(scene, sceneName.name ?? name);
```
Is that proper typescript ? does `.name` exist in sceneName ? Probably not, and thus it happily returns undefined (I cannot say how much I hate languages doing that...) and this explains why all LHCb menus have empty names

9inpachi:
Hi @sponce,

I am sorry about that. I have been a bit busy so not testing the code I write enough. I just pushed a fix that works.

What `processGLTFSceneName` does is return an object with two properties — `name` and `menuNodeName`. It does not return `undefined`. The value is properly processed like it was before.

`VP_>_RFFoil` becomes `menuNodeName=VP` and `name=RFFoil`. What I missed was the usage of nullish coalescing operator (`??`) which falls back only for `null` and `undefined` values not an empty string. So I just needed to pass `undefined` to the function call in LHCb. One other thing I missed was that you also make use of the original `menuNodeName` and append the new path to that. I also added that logic again.

sponce:
Thanks for the quick fix !

PR #429, opened 2022-05-01
Add clipping planes to configuration

### LINKED ISSUE #427: Clipping planes are not modifiable
After wondering why the LHCb Muon would not allow to be clipped while the rest of the detector was fine with it, I found this piece of code : 
```javascript
this.clipPlanes = [
      new Plane(new Vector3(0, 1, 0), 0),
      new Plane(new Vector3(0, -1, 0), 0),
      new Plane(new Vector3(0, 0, 1), -15000),
    ];
```
at https://github.com/HSF/phoenix/blob/master/packages/phoenix-event-display/src/managers/three-manager/index.ts#L93

And we do not seem to have a way to change these hardcoded values. In particular the -15000 is not enough in my case !

9inpachi:
Closes #427

### DIFF
--- a/packages/phoenix-event-display/src/extras/configuration.ts
+++ b/packages/phoenix-event-display/src/extras/configuration.ts
@@ -1,3 +1,4 @@
+import { Plane } from 'three';
 import { PresetView } from './preset-view.model';
 import { EventDataLoader } from '../loaders/event-data-loader';
 import { PhoenixMenuNode } from '../managers/ui-manager/phoenix-menu/phoenix-menu-node';
@@ -22,4 +23,6 @@ export interface Configuration {
   defaultEventFile?: { eventFile: string; eventType: string };
   /** Whether to allow URL options or not (true by default). */
   allowUrlOptions?: boolean;
+  /** Planes for clipping detector geometry. */
+  clippingPlanes?: Plane[];
 }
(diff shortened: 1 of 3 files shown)


sponce:
Works nicely, thanks ! Just one little question out of curiosity and one remark.

The question is about the clipping place that we now change for LHCb (with 15000 replaced with 20000). Why is it necessary at all ? Why not simply drop it ? If it's for performance some how (not looking to far kind of) then I would expect more of these in the other directions. Why is this unique one special ?

Otherwise, the current configuration is a bit misleading in the sense that the user who wants/needs to specify her/his own planes have to know planes 0 and 1 are kind of reserved for internal use. These should be hidden from the user, probably by adding them on the fly in the init method of index.ts

9inpachi:
You have a point. I didn't change any of the values. It has been this way since the initial implementation. I didn't look into how `Plane`s exactly work either.

> Otherwise, the current configuration is a bit misleading in the sense that the user who wants/needs to specify her/his own planes have to know planes 0 and 1 are kind of reserved for internal use. These should be hidden from the user, probably by adding them on the fly in the init method of index.ts

You are correct, I didn't notice this. Let's make it the way you suggested.

sponce:
ok, let me handle it then, as I also want to experiment more with clipping anyway

9inpachi:
Perfect! Thanks.

PR #431, opened 2022-05-06
Improvements in clipping

sponce:
A few improvements have been made to clipping : 
  - the user visible clipping planes are no more polluted by the internal clipping implementation
  - the internal clipping implementation supports clipping up to 360 degrees
  - and the gui has been adapted to be able to use it

There is however still a problem in case of user defined clipping when the clipping angle of the internal clipping in > 180 degrees. In such a case, the internal clipping needs to switch to union mode, which will apply to the user clipping and may result in weird behavior. Now it seems that user clipping was only used by LHCb for unknown reasons and I will soon drop it. So maybe we should rather drop that (just introduced) feature for now.

### DIFF
--- a/packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
+++ b/packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
@@ -23,6 +23,8 @@ import { LoadingManager } from '../loading-manager';
 export class ImportManager {
   /** Planes for clipping geometry. */
   private clipPlanes: Plane[];
+  /** Is clipping intersection or union. */
+  private clipIntersection: boolean;
   /** Object group ID containing event data. */
   private EVENT_DATA_ID: string;
   /** Object group ID containing detector geometries. */
@@ -38,10 +40,12 @@ export class ImportManager {
    */
   constructor(
     clipPlanes: Plane[],
+    clipIntersection: boolean,
     EVENT_DATA_ID: string,
     GEOMETRIES_ID: string
   ) {
     this.clipPlanes = clipPlanes;
+    this.clipIntersection = clipIntersection;
     this.EVENT_DATA_ID = EVENT_DATA_ID;
     this.GEOMETRIES_ID = GEOMETRIES_ID;
     this.loadingManager = new LoadingManager();
@@ -394,7 +398,7 @@ export class ImportManager {
           });
           // Setting up the clipping planes
           child.material.clippingPlanes = this.clipPlanes;
-          child.material.clipIntersection = true;
+          child.material.clipIntersection = this.clipIntersection;
           child.material.clipShadows = false;
         }
       }
--- a/packages/phoenix-event-display/src/managers/three-manager/index.ts
+++ b/packages/phoenix-event-display/src/managers/three-manager/index.ts
@@ -5,6 +5,7 @@ import {
   Vector3,
   Plane,
   Quaternion,
+  Material,
   AmbientLight,
   DirectionalLight,
   AxesHelper,
@@ -73,6 +74,8 @@ export class ThreeManager {
   ];
   /** Clipping planes for clipping geometry. */
   private clipPlanes: Plane[];
+  /** status of cliping intersection */
+  private clipIntersection: boolean;
 
   /**
    * Create the three manager for three.js operations.
@@ -81,6 +84,7 @@ export class ThreeManager {
   constructor(private infoLogger: InfoLogger) {
     this.rendererManager = new RendererManager();
     this.loadingManager = new LoadingManager();
+    this.clipIntersection = true;
   }
 
   /**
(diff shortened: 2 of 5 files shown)


sponce:
Note that I've pushed here on top of several ongoing branches (#425, #419) . They should be first merged and then I'll rebase this.

packages/phoenix-event-display/src/managers/three-manager/index.ts
```diff
@@ -238,23 +243,33 @@ export class ThreeManager {
...
+      const openingAngleQuaternion = new Quaternion();
+      openingAngleQuaternion.setFromAxisAngle(
+          new Vector3(0, 0, 1),
+          ((startingAngle + openingAngle) * Math.PI) / 180
+      );
+      this.clipPlanes[1].normal
+          .set(0, 1, 0)
+          .applyQuaternion(openingAngleQuaternion);
+      const geometries = this.sceneManager.getGeometries();
+      for (const geometry of geometries.children) {
```
9inpachi, line 264: I think going through all the detector geometry objects each time we change the clipping is going to consume A LOT of computing power.

It will be better to have an explicit option to enable clipping intersection manually next to the clipping options.

9inpachi (changes requested):

packages/phoenix-event-display/src/managers/three-manager/index.ts
sponce: To my knowledge (which I admit is till limited), there is no other choice when you want to implement 360 degree clipping. The manual option cannot work as you need to change it automatically when hitting the 180 degree boundary.

sponce: I've just tested it with LHCb full geometry and it seems like it's absolutely fine

packages/phoenix-event-display/src/managers/three-manager/index.ts
```diff
@@ -255,6 +263,16 @@ export class ThreeManager {
     this.clipPlanes[1].normal
       .set(0, 1, 0)
       .applyQuaternion(openingAngleQuaternion);
+    const geometries = this.sceneManager.getGeometries();
```
9inpachi, line 266: One suggestion for improvement is to set and update the `clipIntersection` conditionally like this:

```ts
const isClipIntersectionInvalid =
  (this.clipIntersection && openingAngle > 180) ||
  (!this.clipIntersection && openingAngle < 180);

if (isClipIntersectionInvalid) {
  this.clipIntersection = openingAngle < 180;
  this.sceneManager.getGeometries().traverse((child) => {
    if (child instanceof Mesh) {
      if (child.material instanceof Material) {
        child.material.clipIntersection = this.clipIntersection;
      }
    }
  });
}
```

This only changes the `clipIntersection` once whenever the user goes below or above 180 degrees and doesn't reach the code every time we change the value. The assignment to `this.clipIntersection` at the top of the function is not needed then.

packages/phoenix-event-display/src/managers/three-manager/index.ts
9inpachi: You are correct, it does seem to work fine. One quick comment is that it doesn't seem to work for LHCb although it works fine for ATLAS. The geometry disappears for me for values of opening angle greater than 180 degrees.

9inpachi (changes requested):

packages/phoenix-event-display/src/managers/three-manager/index.ts
sponce: The problem with the (current) LHCb is  what I explain in the description of the MR : collision of this feature with user defined clipping plane (here the z = 20000 one). In such a case, switching to intersection removes everything indeed.
This is why I'm wondering if we should not finally remove this feature that we've just introduced. It's not used by anyone (but LHCb, but by mistake, I will drop it in a MR tomorrow) and collides with the 360 degree clipping.

What do you think ?

packages/phoenix-event-display/src/managers/three-manager/index.ts
sponce: Indeed, that's better than my first implementation. I'll change it.

packages/phoenix-event-display/src/managers/three-manager/index.ts
9inpachi: Ah, I get it now. Thanks for explaining it again. I missed it in the description.

By feature, if you mean defining extra clipping planes through configuration, then I don't have a strong opinion either way. Whatever suits your use case should work.

sponce: Ok, then let's drop these extra planes for now as noone needs them, and we'll come back to this in case.
I'm preparing a new version of this MR

sponce:
@9inpachi this should now be ready, with your changes and the drop of user defined clipping

packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
```diff
@@ -38,10 +40,12 @@ export class ImportManager {
    */
   constructor(
     clipPlanes: Plane[],
+    clipIntersection: boolean,
```
9inpachi, line 43: Do we need this in the import manager? We only ever use it when we change the clipping as far as I understand.

sponce: I think we do in order to apply properly the ongoing clipping to any new geometry we could load. See method `processGeometry`

9inpachi: Thanks for the clarification.

9inpachi (approved):

PR #433, opened 2022-05-10
Allow to specify default opacity in GLTF files

sponce:
This is done via an "opacity" entry in the userData fields of a geometry

### DIFF
--- a/packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
+++ b/packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
@@ -395,6 +395,7 @@ export class ImportManager {
             color: color,
             shininess: 0,
             side: side,
+            opacity: geometry.userData.opacity,
           });
           // Setting up the clipping planes
           child.material.clippingPlanes = this.clipPlanes;


packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
```diff
@@ -245,7 +245,10 @@ export class ImportManager {
                 menuName + geoName.substring(0, pos).replace('_>_', ' > ');
               geoName = geoName.substring(pos + 3);
             }
-            this.processGeometry(geometry, geoName, scale);
+            var opacity = undefined;
+            if ('opacity' in geometry.userData)
+              opacity = geometry.userData['opacity'];
+            this.processGeometry(geometry, geoName, scale, undefined, opacity);
```
9inpachi, line 251: Since the `geometry` is already passed to the function, I don't think we need to do the calculation here. Plus there's a simpler way.

packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
```diff
@@ -392,6 +398,7 @@ export class ImportManager {
             shininess: 0,
             side: side,
           });
+          if (opacity) child['material'].opacity = opacity;
```
9inpachi, line 401: We can get the opacity directly from the geometry right here instead of calculating it in the calling function.

```ts
child.material = new MeshPhongMaterial({
  color: color,
  shininess: 0,
  side: side,
  opacity: geometry.userData.opacity,
});
```

If `geometry.userData.opacity` is `undefined` (when no opacity is set in `userData`), three.js will automatically use a value of `1.0`. But if we want to be explicit about the fallback value, we can also use nullish coalescing `opacity: geometry.userData.opacity ?? 1`.

9inpachi (changes requested):

packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
sponce: :+1:  will change to your version tomorrow

sponce: Ready to go. And on top the change is now a single line ! thanks for the very useful review. I was making it way more complicated than needed !

9inpachi: It seems that the change hasn't been pushed yet.

sponce: oh, I pushed to the wrong repo ! Fixed

9inpachi (approved):

PR #435, opened 2022-05-10
Fixing failing tests, after merge of gltf related PR

sponce:
I'm not sure why it was not visible before, but here is a fix

### DIFF
--- a/packages/phoenix-event-display/src/managers/three-manager/index.ts
+++ b/packages/phoenix-event-display/src/managers/three-manager/index.ts
@@ -342,14 +342,14 @@ export class ThreeManager {
   public loadGLTFGeometry(
     sceneUrl: any,
     name: string,
-    menuNodeName?: string,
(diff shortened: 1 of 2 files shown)


sponce:
@9inpachi I have troubles fixing the tests... On top I do not manage to run them locally. Any idea what goes wrong here ?

9inpachi:
There is only one test failing now.

You should be able to run them locally by navigating to the `phoenix-event-display` package:

```sh
cd packages/phoenix-event-display
yarn test
```

sponce:
@9inpachi I know that should work... but first it fails because I do not have chrome, and if I try to use chromium I have all tests failing with very weird errors. I gave up for the moment

9inpachi:
Oh, I realized this PR and #437 are duplicates that fix the same error in different ways?

sponce:
So let's close this one then

PR #445, opened 2022-05-19
Allow to define a target point in views

sponce:
Target was always 0,0,0 up to now. this applies to default view and all predefined views.
On the way, predefined views for LHCb have been improved

This should only be merged after #443 #442 #444

### DIFF
--- a/packages/phoenix-ng/projects/phoenix-app/src/app/sections/lhcb/lhcb.component.ts
+++ b/packages/phoenix-ng/projects/phoenix-app/src/app/sections/lhcb/lhcb.component.ts
@@ -48,11 +48,33 @@ export class LHCbComponent implements OnInit {
     const configuration: Configuration = {
       eventDataLoader: new PhoenixLoader(),
       presetViews: [
-        new PresetView('Right View', [0, 0, 6000], 'right-cube'),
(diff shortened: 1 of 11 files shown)


packages/phoenix-ng/projects/phoenix-app/src/app/sections/lhcb/lhcb.component.ts
```diff
@@ -48,14 +48,36 @@ export class LHCbComponent implements OnInit {
...
+        new PresetView('Front View', [0, 0, -6000], [0, 0, 0], 'left-cube'),
+        new PresetView(
+          'Top View',
+          [0, 16000, 12000],
+          [0, 0, 12000],
+          'top-cube'
+        ),
       ],
-      defaultView: [-18000, 0, 15000],
+      defaultView: [-8000, 6000, 0, 0, 0, 8000], // x,y,z of position followed by x,y,z of target
```
9inpachi, line 77: The position and target are implicit here. The usage could be confusing for consumers. I think it's better to replace the array with `PresetView` and use it. Your thoughts?

9inpachi (changes requested):

PR #454, opened 2022-05-25
First attempt to cycle through events

### LINKED ISSUE #447: Having a button to rotate on events
It would be nice that one can activate a mode where events are automatically rotated every x seconds.

sponce:
At this stage the cycling speed is hardcoded (5s between events)
And the activation of cycling is hidden in the list of events by adding a last item celles "cycle through"

This is basically a proof of concept to start the discussion

Closes #447 

Should only be merged after #453

### DIFF
--- a/packages/phoenix-event-display/src/loaders/phoenix-loader.ts
+++ b/packages/phoenix-event-display/src/loaders/phoenix-loader.ts
@@ -270,6 +270,7 @@ export class PhoenixLoader implements EventDataLoader {
         scalePlanarCaloCells
       );
 
+      const collections = {};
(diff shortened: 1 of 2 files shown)


9inpachi:
I was thinking in the direction of adding a new toggle in the icon bar. It would be a component that can take the cycling speed as an input.

9inpachi:
I added the component to this new branch: https://github.com/HSF/phoenix/tree/feat-cycle-events

There is no icon or tests yet. You can cherry pick or continue on that branch if you want. The component itself should work fine.

sponce:
superseeded by #455, so closing

PR #455, opened 2022-05-26
Component to cycle through events

### LINKED ISSUE #447: Having a button to rotate on events

9inpachi:
Closes #447


9inpachi:
@sponce I haven't added it to the UI menu yet. I am not sure if it should be only for LHCb or all experiments.

If you want to add it only to LHCb, you can append it to the UI menu in `lhcb.component.html` template.

```html
<app-ui-menu>
  <app-cycle-events [interval]="5000"></app-cycle-events>
</app-ui-menu>
```

PR #460, opened 2022-06-20
Changes in LHCb data and gui

sponce:
Not to be merged for now, as the gui is not optimal (cycle through events button not in the right place)


9inpachi:
I have added a new component that can be easily used to only define UI menu items needed. See https://github.com/HSF/phoenix/pull/461.

sponce:
Thanks a lot @9inpachi. I've used it already and it works nicely. I believe this PR can thus be merged now.

PR #462, opened 2022-06-21
Add unit tests for lib and helpers

DamianArado:
Code coverage would be low as of now and will be improved in due time. Also, I've noticed that testing static methods doesn't increase coverage at all. Need to find a workaround for that as well.


DamianArado:
Errors are thrown in unit tests for `phoenix-app` and `phoenix-ui-components` due to missing `@types/jasmine` which was removed by me earlier as it was giving rise to type conflicts. There is a possibility that we migrate our unit tests of `phoenix-app` and `phoenix-ui-components` to Jest as well. So, I'll take care of that.

9inpachi (changes requested):
Thanks for the initial batch of tests. :)

I think we need to synchronize on how the tests should be written so everything is consistent. I have the following comments on the tests you have written so far.

* We should either use `test` or `it` for test cases across the entire repo to keep everything consistent. If you have a preference for `test` and plan to update all the other tests in the repo, then it's fine to use it. I see a usage of `it` in your test as well so I would recommend to use `test` for it as well if you plan to do it that way. IMO, `it` is more readable since it makes a human-readable sentence like `it should get data from browser local storage` and is closer to TDD. But it's a matter of preference.
* Please keep in mind that unit tests are _not_ written to reach a 90%+ coverage. They are written to test behavior of a unit (class, function etc.). I notice a lot of focus on increasing coverage wherever possible which I think is the wrong approach. If we test the behaviors correctly, we should automatically have a complete coverage of the code.
* Constants are not tested separately. Testing the code that uses those constants should ideally cover them.
* Most of the test cases I think can be improved so they test the behavior instead of just covering a piece of code. This means that we should test "what" the function does rather than "how" it does it and it should be clear in the test case description what's being tested.
  For example:
  **Bad:** `test('getFromLocalStorage')`
  **Better:** `it('should get existing data from browser local storage')`
* The code we are supposed to test should not be mocked. A method/object should only be mocked if it is not a part of the test case and the code logic that we are testing. Or if we need it to produce a different output to test our logic (for example, spying and mocking the `window` object).
* We should avoid having assertions that have nothing to do with the test case and code to be tested. For example, the test case `'should not call callback if not subscribed'` is redundant since it doesn't actually test anything.
* Types are not tested. They don't contain any logic that has to be tested.

Would be great if you could update the tests according to these comments. :)

DamianArado:
Thanks a lot for the feedback @9inpachi 🙂

> IMO, it is more readable since it makes a human-readable sentence like it should get data from browser local storage and is closer to TDD   

I agree. I will make sure to use `it` in each test case.  

> If we test the behaviors correctly, we should automatically have a complete coverage of the code  

Yes, I'll keep this in mind from now. 😅

> Constants are not tested separately. Testing the code that uses those constants should ideally cover them.  

I see.  

> we should test "what" the function does rather than "how" it does it and it should be clear in the test case description what's being tested   

Okay, I'll look into it and make the changes wherever necessary.   

> The code we are supposed to test should not be mocked. A method/object should only be mocked if it is not a part of the test case and the code logic that we are testing.  

How can I test the behavior of static methods? 😅 Since, I can't call a static method directly using the object of that class, I think mocking is the only option left. Or is there any other way as well that I'm missing?  

>  For example, the test case 'should not call callback if not subscribed' is redundant since it doesn't actually test anything.  

I see, I'll remove that.  

> Types are not tested. They don't contain any logic that has to be tested

Sure, removing that.

9inpachi:
> How can I test the behavior of static methods? sweat_smile Since, I can't call a static method directly using the object of that class, I think mocking is the only option left. Or is there any other way as well that I'm missing?

Just call them directly. Think of them as standalone functions.

Here is an example:

source-class.ts
```ts
class SourceClass {
  public static greeting(): string {
    return 'Hello there';
  }
}
```

source-class.test.ts
```ts
describe('SourceClass', () => {
  it('should give a greeting', () => {
    expect(SourceClass.greeting()).toBe('Hello there');
  });
});
```

DamianArado:
@9inpachi Please check whenever you have time and let me know if anything else needs to be improved/changed. After this, I'll be moving on to `managers`.

packages/phoenix-event-display/src/tests/lib/models/cut.model.test.ts
```diff
@@ -0,0 +1,23 @@
...
+  });
+
+  it('should reset the values', () => {
+    model.reset();
+    expect(model.minValue).toBe(0);
+    expect(model.maxValue).toBe(1);
+  });
+
+  it('can be created with default values', () => {
+    expect(model).toBeTruthy();
```
9inpachi, line 21: This is the same as the first test case. Do we need to test something different here?

packages/phoenix-event-display/src/tests/lib/models/cut.model.test.ts
```diff
@@ -0,0 +1,23 @@
...
+  beforeEach(() => {
+    model = new Cut('test', 0, 1, 1);
+  });
+
+  it('should create an instance of Cut model', () => {
+    expect(model).toBeTruthy();
+  });
+
+  it('should reset the values', () => {
+    model.reset();
```
9inpachi, line 15: The values will be reset by default. What we can do here is first assign the values and then reset them.

```ts
it('should reset the values', () => {
  model.minValue = 10;
  model.maxValue = 20;

  model.reset();

  expect(model.minValue).toBe(0);
  expect(model.maxValue).toBe(1);
});
```

packages/phoenix-event-display/src/tests/managers/three-manager/xr/xr-manager.test.ts
```diff
@@ -1,37 +1,32 @@
...
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('XRManager should be created', () => {
+    expect(xrManager).toBeTruthy();
+  });
+
+  it('It should get the group containing the camera for XR', () => {
```
9inpachi, line 18: NIT: `it('should get the group containing the camera for XR', () => {});`

packages/phoenix-event-display/src/tests/managers/three-manager/xr/xr-manager.test.ts
```diff
@@ -1,37 +1,32 @@
...
+    expect(xrManager.getCameraGroup()).toBe(xrManager.cameraGroup);
+    const xrcamera = new Camera();
+    xrManager.xrCamera = xrcamera;
+    expect(xrManager.getXRCamera()).toBe(xrcamera);
+    const cameraPosition = new Vector3(0, 0, 0.1);
+    xrManager.cameraGroup.position.copy(cameraPosition);
+    expect(xrManager.getCameraGroup().position).toEqual(cameraPosition);
+  });
+
+  it('It should end the current XR session', () => {
```
9inpachi, line 29: NIT: `it('It should end the current XR session', () => {});`

9inpachi (changes requested):
Good job! Looks great now.

I just added some small comments.

packages/phoenix-event-display/src/tests/lib/models/cut.model.test.ts
DamianArado: missed that. 😅

packages/phoenix-event-display/src/tests/lib/models/cut.model.test.ts
DamianArado: Sure, changing...

packages/phoenix-event-display/src/tests/managers/three-manager/xr/xr-manager.test.ts
DamianArado: This will be updated along with the unit tests for `managers`.

packages/phoenix-event-display/src/tests/managers/three-manager/xr/xr-manager.test.ts
DamianArado: This will be updated along with the unit tests for `managers`.

9inpachi (approved):
Thanks. Great job on the first batch!

Issue #457, opened 2022-06-09
Rewrite unit tests of phoenix using Jest


DamianArado:
As agreed upon with @9inpachi and @EdwardMoyse, this change would ensure quicker execution of tests for the `phoenix-event-display` library.  

Current execution time of tests (checked locally): https://pastebin.com/CWTEnRy5

DamianArado:
All the unit tests in phoenix should now use Jest instead of Jasmine and Karma. We will use Cypress for e2e tests in the future, so there is no dependency on Jasmine as was the case with Protractor.

PR #473, opened 2022-07-02
test(event-display): update unit test for XRManager

DamianArado:
Please let me know if any test case needs improvement among the ones added in this PR.

This PR will precede a PR that will try and improve unit test cases in individual files for the whole of the `phoenix-event-display` module since the testing suite for `phoenix-event-display` is almost complete. This whole testing suite is written primarily keeping in mind to test behaviors correctly and to be easily maintainable by developers in the future.

Also, `three-manager/index.ts` and `ui-manager/index.ts` have `WebGLRenderer` as one of their dependency so that's messing things in their tests as when I mock these classes, by default, their objects are `undefined`. Need to find a way to test these two.


packages/phoenix-event-display/src/tests/loaders/phoenix-loader.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    const infoLogger = new InfoLogger();
+
+    jest.spyOn(phoenixLoader, 'buildEventData');
+    phoenixLoader.buildEventData(
+      eventData,
+      graphicsLibrary,
+      uiManager,
+      infoLogger
+    );
+    expect(phoenixLoader.buildEventData).toHaveBeenCalled();
```
9inpachi, line 35: I think this assertion is redundant. We are calling the function in the previous statement and just checking here if it's called. We should test if it actually loads the object types specified in the event data.

9inpachi: I have the same comments for all the other test cases in this suite. We are not testing anything here. We need to assert that the correct methods are called internally in these functions.

And if the functions return anything, that return value should be asserted.

packages/phoenix-event-display/src/tests/loaders/script-loader.test.ts
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  it('should load a script dynamically from a URL', () => {
+    const scriptURL = 'https://www.google.com/jsapi';
+    const scriptFor = 'google_jsapi';
+    const parentElement = document.getElementsByTagName('head')[0];
+
+    jest.spyOn(ScriptLoader, 'loadScript');
+    ScriptLoader.loadScript(scriptURL, scriptFor, parentElement);
+    expect(ScriptLoader.loadScript).toHaveBeenCalledWith(
```
9inpachi, line 24: This also just asserts if the previous statement executed. If you call a function, it is going to be called, there is no need to assert that here.

What we need to assert here is that the script is correctly loaded in the DOM.

packages/phoenix-event-display/src/tests/loaders/trackml-loader.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  it('should create an instance', () => {
+    expect(trackmlLoader).toBeTruthy();
+  });
+
+  it('should process hits', () => {
+    const hits = `1,0,0,0,0,0,0
+    2,0,0,0,0,0,0`;
+    jest.spyOn(trackmlLoader, 'processHits');
+    trackmlLoader.processHits(hits);
+    expect(trackmlLoader.processHits).toHaveBeenCalledWith(hits);
```
9inpachi, line 22: Same comment here. All these test cases don't test anything.

packages/phoenix-event-display/src/tests/managers/loading-manager.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  it('should add an item which is to be loaded', () => {
+    loadingManager.addLoadableItem('item');
+    expect(loadingManager.toLoad.length).toBe(1);
+  });
+
+  it('should call the function when an item has finished loading', () => {
+    jest.spyOn(loadingManager, 'itemLoaded');
+    loadingManager.addLoadableItem('item');
+    loadingManager.itemLoaded('item');
+    expect(loadingManager.itemLoaded).toHaveBeenCalled();
```
9inpachi, line 27: This is redundant. The correct way to assert that the loading was finished would be to add a load listener (`jest.fn()`) using `addLoadListener` and make sure that the load listener/callback was called.

packages/phoenix-event-display/src/tests/managers/loading-manager.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  it('should call the function when loading of an item progresses', () => {
+    jest.spyOn(loadingManager, 'onProgress');
+    loadingManager.addLoadableItem('item');
+    loadingManager.onProgress('item', 50);
+    expect(loadingManager.onProgress).toHaveBeenCalled();
```
9inpachi, line 34: Same here. We need to add a callback/progress listener and see if it was called when the progress changed.

packages/phoenix-event-display/src/tests/managers/loading-manager.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  it('should add a listener for when all items have loaded', () => {
+    jest.spyOn(loadingManager, 'addLoadListener');
+    loadingManager.addLoadListener(() => {});
+    expect(loadingManager.addLoadListener).toHaveBeenCalled();
```
9inpachi, line 40: All of the remaining test cases in this suite including this one shouldn't be needed. If you are testing the behavior you check if the listeners were correctly called when the loading of all items was completed. Which I suggested in my previous comments.

packages/phoenix-event-display/src/tests/managers/loading-manager.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  it('should add a listener for when all items have loaded with check if there', () => {
+    jest.spyOn(loadingManager, 'addLoadListenerWithCheck');
+    loadingManager.addLoadListenerWithCheck(() => {});
+    expect(loadingManager.addLoadListenerWithCheck).toHaveBeenCalled();
```
9inpachi, line 46: This is a special case, so we will need a different test case for this where we have the loading manager in a state where all items have loaded and then check if the callback/listener is immediately called even if we add the listener after the items have loaded.

packages/phoenix-event-display/src/tests/managers/loading-manager.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  it('should add a callback to listen when the progress of an item increases', () => {
+    jest.spyOn(loadingManager, 'addProgressListener');
+    loadingManager.addProgressListener(() => {});
+    expect(loadingManager.addProgressListener).toHaveBeenCalled();
```
9inpachi, line 52: This test case is redundant.

9inpachi (changes requested):
I have seen half of the code. Sorry, I am not continuing the review here because there are a lot of test cases and most of them follow the same pattern.

1. Spy on a method that is to be tested.
2. Call the method.
3. Assert that the method was called.

This is redundant since what we need to test is that the internals of the function are working correctly. We know for a fact that we called the function in the previous statement, so the assertion in the next statement doesn't make sense.

If the function returns something, we assert that it returns the correct value. Otherwise, we assert that things that are interlinked inside the object/function are called/initiated correctly. Like how I explained it for the listeners and functions in loading manager.

I honestly feel that this is the same as testing for coverage.

By "assert", I mean "expect".

packages/phoenix-event-display/src/tests/managers/loading-manager.test.ts
9inpachi: Here's how I would test it:

```ts
it('should call the listeners when all items have loaded', () => {
  const callback = jest.fn();
  loadManager.addLoadListener(callback);
  loadingManager.addLoadableItem('item');
  loadingManager.itemLoaded('item');
  expect(callback).toHaveBeenCalled();
});
```

DamianArado:
@9inpachi, thanks for correcting me. I admit I am on the wrong approach to using spies in many test cases as I saw some of the tests written earlier had used spies too. But, instead, we need to test the internals of the methods correctly as you said. I'll keep this in mind.

Let me correct it one test suite - at a time. I'll let you know as soon as I'm done. I really appreciate the time you spend reviewing the code. 🙌

packages/phoenix-event-display/src/tests/managers/loading-manager.test.ts
DamianArado: Thanks! Right ✅

packages/phoenix-event-display/src/tests/loaders/objects/cms-objects.test.ts
```diff
@@ -0,0 +1,29 @@
...
+      front_3: [1, 1, 0],
+      front_4: [0, 1, 0],
+      back_1: [0, 0, 1],
+      back_2: [1, 0, 1],
+      back_3: [1, 1, 1],
+      back_4: [0, 1, 1],
+    };
+
+    const muonChamber = CMSObjects.getMuonChamber(muonChamberParams);
+    expect(muonChamber).toBeTruthy();
```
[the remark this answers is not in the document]
DamianArado: very helpful.

9inpachi:
Yes, please do. LGTM.

packages/phoenix-event-display/src/tests/loaders/objects/phoenix-objects.test.ts
```diff
@@ -0,0 +1,114 @@
...
+  });
+
+  it('should process the Hits from the given parameters and get them as a geometry', () => {
+    const hitParams = {
+      pos: [0, 0, 0],
+      dparams: [1, 0, 0],
+      indices: [0, 1, 2],
+    };
+
+    const hit = PhoenixObjects.getHits(hitParams);
```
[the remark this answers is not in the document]
DamianArado: I'm trying to use `hitsParams` as -
```
const hitsParams = {
      pos: [
        [
          -2446.784912109375, -2425.1064453125, 7826.09912109375,
          -2446.784912109375, -1.1222461462020874, 7826.09912109375,
        ],
        [
          -2448.534912109375, -2425.1064453125, 7826.09912109375,
          -2448.534912109375, -1.1222461462020874, 7826.09912109375,
        ],
        [
          -2446.784912109375, -2425.1064453125, 7826.09912109375,
          -2446.784912109375, -1.1222461462020874, 7826.09912109375,
        ],
      ],
    };
```

but getting `No hits! Aborting from getHits` in console due to -
```
if (hitsParams.length > 0) {
--- logic ----
} else {
      console.log('No hits! Aborting from getHits.');
      return new Object3D();
}
```

I took these as `pos` from the FTHits collection here: https://hepsoftwarefoundation.org/phoenix/#/lhcb. Can I please get an example of hits that can be used?

packages/phoenix-event-display/src/tests/loaders/objects/phoenix-objects.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      ],
+      dparams: [[1, 0, 0, 3]],
+      indices: [0, 1, 2],
+    };
+
+    const track = PhoenixObjects.getTrack(trackParams);
+    expect(track).toBeTruthy();
+  });
+
+  it('should process the Jet from the given parameters and get it as a geometry', () => {
```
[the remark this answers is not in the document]
9inpachi: Yes, please do. LGTM.

packages/phoenix-event-display/src/tests/loaders/objects/phoenix-objects.test.ts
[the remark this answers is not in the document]
9inpachi: You will find the structure of most of the event data objects here: <https://github.com/HSF/phoenix/tree/master/packages/phoenix-ng/projects/phoenix-app/src/assets/files/event_data>

The Phoenix JSON format of event data is also documented here on a higher level: <https://github.com/HSF/phoenix/blob/master/guides/users.md#event-data>

Hope this helps.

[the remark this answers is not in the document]
9inpachi: FTHits are of the following format:

```json
{
  "EventRun": {
    "Hits": {
      "HitsCollection/FTHits": [
        {
          "type": "Line"
          "pos": [2, 10, 30, 40]
        }
      ]
    }
  }
}
```

You can check the event data here for the format: <https://github.com/HSF/phoenix/blob/656bcb8e1a98122bd52afcee940a72a14fc8a7c7/packages/phoenix-ng/projects/phoenix-app/src/assets/files/lhcb/LHCbEventData.json>

This creates a `Line` type hit as specified in the event data itself. See <https://github.com/HSF/phoenix/blob/656bcb8e1a98122bd52afcee940a72a14fc8a7c7/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts#L262>

> I'm trying to use hitsParams as

The `hitsParams` you are trying to use don't have the correct format. `hitsParams` is expected to be an array of objects. The correct format would be:

```json
{
  "EventRun": {
    "Hits": {
      "HitsCollection": [
        [2, 10, 30],
        [2, 10, 30]
      ]
    }
  }
}
```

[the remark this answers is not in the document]
9inpachi: In general, you can copy paste the simplest data from event data files we have in the assets: <https://github.com/HSF/phoenix/tree/656bcb8e1a98122bd52afcee940a72a14fc8a7c7/packages/phoenix-ng/projects/phoenix-app/src/assets/files>

Some of them will have the "Hits" data type and under it custom named collections (I shared a link to the users manual which should help you understand what I mean). You can check these event data to understand the formats `hitsParams` should be in.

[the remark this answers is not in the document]
DamianArado: Thanks a lot!

DamianArado:
@9inpachi, please take a look whenever you have time and if any other changes are required, do let me know.

packages/phoenix-event-display/src/tests/managers/three-manager/effects-manager.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  it('should create an instance', () => {
+    expect(effectsManager).toBeTruthy();
+  });
+
+  it('should initialize the outline pass for highlighting hovered over event display elements', () => {
+    const outlinePass = effectsManager.addOutlinePassForSelection();
+
+    expect(outlinePass).toBeInstanceOf(OutlinePass);
+    expect(outlinePass.selectedObjects.length).toBe(0);
```
9inpachi, line 41: I don't think there is a need to check all these properties. We just need to assert if the composer contains the outline pass after we add it to the composer.

packages/phoenix-event-display/src/tests/managers/three-manager/scene-manager.test.ts
```diff
@@ -0,0 +1,235 @@
...
+    new AxesHelper().type,
+  ];
+
+  beforeEach(() => {
+    sceneManager = new SceneManager(ignoreList, false);
+  });
+
+  it('should update position of directional light for each frame rendered', () => {
+    sceneManager.updateLights(new Camera());
+    expect(sceneManager.cameraLight).toBeUndefined();
```
9inpachi, line 34: This does not test the functionality correctly. We need to make sure that the `cameraLight` has the position of `camera`.

packages/phoenix-event-display/src/tests/managers/three-manager/scene-manager.test.ts
```diff
@@ -0,0 +1,235 @@
...
+  });
+
+  it('should get a clean copy of the scene', () => {
+    const scene = new Group();
+    scene.add(new Mesh(new BoxGeometry(), new MeshBasicMaterial()));
+    const cleanScene = sceneManager.getCleanScene();
+
+    expect(cleanScene).toBeInstanceOf(Scene);
+    expect(cleanScene.children.length).toBe(0);
```
9inpachi, line 43: We need to check here that the clean scene doesn't contain the ignored objects type. No need to check for `up`, `position` and `scale` properties.

packages/phoenix-event-display/src/tests/managers/three-manager/scene-manager.test.ts
```diff
@@ -0,0 +1,235 @@
...
+    expect(eventGroup.parent.name).toBe('EventData');
+    expect(eventGroup.parent.parent).toBeInstanceOf(Scene);
+    expect(eventGroup.parent.parent.children.length).toBe(6);
+    expect(eventGroup.userData).toEqual({});
+  });
+
+  it('should clear event data of the scene', () => {
+    jest.spyOn(sceneManager, 'getEventData');
+    sceneManager.clearEventData();
+    expect(sceneManager.getEventData).toHaveBeenCalledTimes(2);
```
9inpachi, line 77: I think we should instead assert there that the scene does not contain the event data group. The assertion right now does not test the intended behavior.

packages/phoenix-event-display/src/tests/managers/ui-manager/phoenix-menu/phoenix-menu-ui.test.ts
```diff
@@ -0,0 +1,124 @@
...
+  it('should create an instance', () => {
+    expect(phoenixMenuUI).toBeTruthy();
+  });
+
+  it('should add geometry (detector geometry) folder to the menu', () => {
+    phoenixMenuUI.addGeometryFolder();
+
+    expect(phoenixMenuUIPrivate.geomFolder.name).toBe('Detector');
+    expect(phoenixMenuUIPrivate.geomFolder).toBeInstanceOf(PhoenixMenuNode);
+    expect(phoenixMenuUIPrivate.geomFolder.toggleState).toBeTruthy();
```
9inpachi, line 38: Just assert what we expect as we mention it in the test case statement. We don't need to check each property of the object. This is true for all the other tests you wrote as well.

packages/phoenix-event-display/src/tests/managers/ui-manager/phoenix-menu/phoenix-menu-ui.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    expect(phoenixMenuUIPrivate.labelsFolder.configs[2].type).toBe('button');
+    expect(phoenixMenuUIPrivate.labelsFolder.parent).toBeInstanceOf(
+      PhoenixMenuNode
+    );
+    expect(phoenixMenuUIPrivate.labelsFolder.parent.name).toBe('test');
+  });
+
+  it('should remove label from the menu and scene if it exists', () => {
+    phoenixMenuUI.addLabelsFolder('test');
+    phoenixMenuUI.removeLabel('test');
```
9inpachi, line 74: This does not test the intended behavior. Pass in a mock label node of type `PhoenixMenuNode` and make sure that it's `remove` method was called.

packages/phoenix-event-display/src/tests/managers/url-options-manager.test.ts
```diff
@@ -0,0 +1,27 @@
...
+  });
+
+  it('should create an instance', () => {
+    expect(urlOptionsManager).toBeTruthy();
+  });
+
+  it('should get options from URL set through query parameters', () => {
+    expect(urlOptionsManager.getURLOptions()).toBeInstanceOf(URLSearchParams);
+  });
+});
```
9inpachi, line 27: There are other main public methods that seem to be not tested?

9inpachi (changes requested):

packages/phoenix-event-display/src/tests/managers/three-manager/xr/xr-manager.test.ts
```diff
@@ -1,32 +1,48 @@
...
+  it('should create an instance of XRManager', () => {
+    expect(xrManager).toBeTruthy();
   });
 
-  it('XRManager should be created', () => {
-    expect(xrManager).toBeTruthy();
+  it('should set and configure the XR session', () => {
+    const onSessionStarted = () => {};
+    const onSessionEnded = () => {};
+    xrManager.setXRSession(renderer, onSessionStarted, onSessionEnded);
```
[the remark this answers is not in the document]
DamianArado: For me, in my IDE, `navigator.xr` is undefined as `WebXR` is not available. I hence, cannot spy on `requestSession()`. 

I also tried using `(navigator as any)?.xr`, but its still `undefined`.

Is there any other way? 🤔

packages/phoenix-event-display/src/tests/managers/three-manager/xr/xr-manager.test.ts
```diff
@@ -1,32 +1,48 @@
...
+
+    const cameraPosition = XRSessionType.AR
+      ? xrcamera.position
+      : new Vector3(0, 0, 0.1);
     xrManager.cameraGroup.position.copy(cameraPosition);
+    xrManager.cameraGroup.add(xrcamera);
+
+    expect(xrManager.getCameraGroup()).toEqual(xrManager.cameraGroup);
     expect(xrManager.getCameraGroup().position).toEqual(cameraPosition);
+    expect(xrManager.getCameraGroup().children[0]).toEqual(xrcamera);
```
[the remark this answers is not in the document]
DamianArado: Gotcha! 🙌

packages/phoenix-event-display/src/tests/managers/url-options-manager.test.ts
DamianArado: Sure, done. ✅

Just this one remains due to this line being executed -
https://github.com/HSF/phoenix/blob/50d629e7dbf4432e3f1dd8b9a49466b53ecc0fc8/packages/phoenix-event-display/src/managers/url-options-manager.ts#L62

packages/phoenix-event-display/src/tests/loaders/objects/cms-objects.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    const mesh = muonChamber.children[0] as Mesh;
+    expect(mesh.position).toBeInstanceOf(Vector3);
+    expect(mesh.rotation).toBeInstanceOf(Euler);
+    expect(mesh.geometry.attributes.position.count).toBe(36);
+    expect(mesh.receiveShadow).toBe(false);
+    expect(mesh.frustumCulled).toBe(true);
+    expect(mesh.renderOrder).toBe(0);
+    expect(mesh.animations).toHaveLength(0);
+    expect(mesh.geometry.type).toBe('BufferGeometry');
+    expect(mesh.material).toBeInstanceOf(MeshBasicMaterial);
```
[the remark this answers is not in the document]
DamianArado: Right! too many assertions can be an overkill, need to remove them in other files as well.

packages/phoenix-event-display/src/tests/managers/three-manager/xr/xr-manager.test.ts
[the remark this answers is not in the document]
9inpachi: Try if `window.navigator` works. If not, you can try declaring `navigator` by creating a `global.d.ts` file in `tests/` directory to declare it as a global type.

```ts
declare global {
    interface Window { navigator: any; }
}
```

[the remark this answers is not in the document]
DamianArado: Same with `window.navigator`. When I created the file inside /tests with contents -
```
export {};

declare global {
    interface Window {
      navigator: any;
    }
}

```

It says `All declarations of 'navigator' must have identical modifiers.`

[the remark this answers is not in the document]
9inpachi: I tried it out and the test can be written this way:

```ts
it('should set and configure the XR session', () => {
  jest
    .spyOn(xrManager as any, 'onXRSessionStarted')
    .mockImplementation(jest.fn());
  const onSessionStarted = jest.fn();
  const onSessionEnded = jest.fn();
  const requestSessionSpy = jest.fn().mockResolvedValue(Promise.resolve({}));

  Object.defineProperty(window.navigator, 'xr', {
    value: {
      requestSession: requestSessionSpy,
    },
  });

  xrManager.setXRSession(renderer, onSessionStarted, onSessionEnded);
  expect(requestSessionSpy).toHaveBeenCalled();
});
```

packages/phoenix-event-display/src/tests/managers/state-manager.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  beforeEach(() => {
+    stateManager = new StateManager();
+  });
+
+  it('should create an instance', () => {
+    expect(stateManager).toBeTruthy();
+  });
+
+  it('should get the instance of stateManager', () => {
+    expect(StateManager.getInstance()).toBeTruthy();
```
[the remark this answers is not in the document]
DamianArado: sure 💯

packages/phoenix-event-display/src/tests/loaders/objects/phoenix-objects.test.ts
```diff
@@ -0,0 +1,440 @@
...
+    expect(planarCaloCellObject.up.z).toBe(0);
+
+    expect(planarCaloCellObject.position).toBeInstanceOf(Vector3);
+    expect(planarCaloCellObject.position.x).toBe(0);
+    expect(planarCaloCellObject.position.y).toBe(0);
+    expect(planarCaloCellObject.position.z).toBe(0);
+
+    expect(planarCaloCellObject.rotation).toBeInstanceOf(Euler);
+
+    expect(planarCaloCellObject.userData.energy).toBe(
```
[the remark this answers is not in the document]
9inpachi: `userData` is a subset of `trackParams`. So the order will be reversed.

```ts
expect(trackParams).toMatchObject(userData);
```

packages/phoenix-event-display/src/tests/managers/three-manager/xr/xr-manager.test.ts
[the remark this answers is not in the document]
DamianArado: Thanks a lot!

packages/phoenix-event-display/src/tests/loaders/objects/phoenix-objects.test.ts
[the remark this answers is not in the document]
DamianArado: neat!

packages/phoenix-event-display/src/tests/managers/url-options-manager.test.ts
DamianArado: This is the main issue: `Cannot read properties of undefined (reading 'addLoadableItem')`

it's just that this method is not callable

https://github.com/HSF/phoenix/blob/50d629e7dbf4432e3f1dd8b9a49466b53ecc0fc8/packages/phoenix-event-display/src/managers/url-options-manager.ts#L79

9inpachi: That's because you are mocking the event display import I think. You can either mock these functions of the event display or use the event display as is (if that works).

DamianArado: I can't use EventDisplay as it is without mocking due to this dependency hierarchy: `EventDisplay <- UIManager <- ThreeManager <- WebGLRenderer`. By default, `eventDisplay` is undefined due to this.

9inpachi:
Then like I mentioned, you need to mock the import and define mock methods
for the ones that are called in the URL options manager. Check Jest docs
for how to mock an import.

packages/phoenix-event-display/src/tests/managers/state-manager.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    stateManager.setCamera(camera);
+    expect(stateManager.activeCamera).toBe(camera);
+  });
+
+  it('should set the event display for state', () => {
+    const eventDisplay = new EventDisplay();
+    stateManager.setEventDisplay(eventDisplay);
+    expect(stateManager.eventDisplay).toBe(eventDisplay);
+  });
+});
```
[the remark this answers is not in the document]
DamianArado: ✅ done

packages/phoenix-event-display/src/tests/managers/ui-manager/phoenix-menu/phoenix-menu-ui.test.ts
DamianArado: Done ✅. Thanks!

packages/phoenix-event-display/src/tests/loaders/phoenix-loader.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    const infoLogger = new InfoLogger();
+
+    jest.spyOn(infoLogger, 'add');
+    phoenixLoader.buildEventData(
+      eventData,
+      graphicsLibrary,
+      uiManager,
+      infoLogger
+    );
+    expect(infoLogger.add).toBeCalledTimes(1);
```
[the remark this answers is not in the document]
DamianArado: Thank you, it's very helpful  ^_^

packages/phoenix-event-display/src/tests/loaders/phoenix-loader.test.ts
```diff
@@ -0,0 +1,109 @@
...
+  it('should create an instance', () => {
+    expect(phoenixLoader).toBeTruthy();
+  });
+
+  it('should take an object that represents one event and add the different objects to the graphics library and the UI controls', () => {
+    const eventData = [
+      {
+        'event number': 120,
+        'run number': 111,
+        pos: [
```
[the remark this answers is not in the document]
DamianArado: Thanks a lot! ^_^

packages/phoenix-event-display/src/tests/loaders/script-loader.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    const scriptURL = 'https://www.google.com/jsapi';
+    const scriptFor = 'google_jsapi';
+    const parentElement = document.getElementsByTagName('head')[0];
+
+    const promise = ScriptLoader.loadScript(
+      scriptURL,
+      scriptFor,
+      parentElement
+    );
+    expect(promise).toBeInstanceOf(Promise);
```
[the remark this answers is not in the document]
DamianArado: Yes, we can query the DOM but I tried to await for more than 50s, it's not going beyond this callback -
https://github.com/HSF/phoenix/blob/50d629e7dbf4432e3f1dd8b9a49466b53ecc0fc8/packages/phoenix-event-display/src/loaders/script-loader.ts#L68

Hence, moving on with the latter approach. We can also do check using `document.scripts`.

DamianArado:
@9inpachi, I think we don't need any more PRs for `phoenix-event-display` as of now. According to me, only the tests for `three-manager/index` and `ui-manager/index` are left to be written (due to the limitations of the `jsdom` environment in Jest) which I'll keep as a TODO for myself. The merging of this PR essentially closes: https://github.com/HSF/phoenix/issues/457.

Apart from that, I hope these unit tests will be easily maintained by me or any other developer working on Phoenix, in the future. Still, If something needs to change, please let me know. 

After this, I'm planning to change the unit tests for `phoenix-ui-components` and `phoenix-app` to use Jest too so that there is uniformity in the testing setup of both the library and the app. Firstly, I'll try changing the setup in `phoenix-app` and see how it goes.

packages/phoenix-event-display/src/tests/managers/ui-manager/dat-gui-ui.test.ts
```diff
@@ -0,0 +1,79 @@
...
+      datGUIMenuPrivate.eventFolder.__folders['Event Data']
+    ).toBeInstanceOf(GUI);
+  });
+
+  it('should add collection folder and its configurable options to the event data type (tracks, hits etc.) folder', () => {
+    datGUIMenu.addEventDataFolder();
+    datGUIMenu.addCollection('Event Data', 'test');
+
+    expect(
+      datGUIMenuPrivate.eventFolder.__folders['Event Data']
```
9inpachi, line 68: It seems this doesn't check if the collection folder was added or not.

packages/phoenix-event-display/src/tests/managers/ui-manager/phoenix-menu/phoenix-menu-ui.test.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  it('should add collection folder and its configurable options to the event data type (tracks, hits etc.) folder', () => {
+    phoenixMenuUI.addEventDataFolder();
+    phoenixMenuUI.addEventDataTypeFolder('test');
+
+    jest.spyOn(three, 'getColorManager');
+
+    phoenixMenuUI.addCollection('test', 'Tracks');
+
+    expect(three.getColorManager).toHaveBeenCalled();
```
9inpachi, line 93: This doesn't check if the collection folder/node was added.

https://github.com/HSF/phoenix/blob/master/packages/phoenix-event-display/src/managers/ui-manager/phoenix-menu/phoenix-menu-ui.ts#L222

packages/phoenix-event-display/src/tests/loaders/phoenix-loader.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      'run number',
+      'Hits',
+      'event number',
+      'run number',
+      'Hits',
+    ]);
+  });
+
+  it('should get the collection with the given collection name from the event data', () => {
+    const collectionName = {
```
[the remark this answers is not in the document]
DamianArado: Yes ✅

packages/phoenix-event-display/src/tests/managers/ui-manager/dat-gui-ui.test.ts
DamianArado: It will be added but yeah this test case is not explicit enough. Removing it.

packages/phoenix-event-display/src/tests/managers/ui-manager/phoenix-menu/phoenix-menu-ui.test.ts
DamianArado: I wanted to spy on typeFolder.addChild but I don't understand why it does not execute further from here in this function where we find out the collectionNode - 
https://github.com/HSF/phoenix/blob/5ec4ed3fb8818144530e56a5e5fcb4c46ec54557/packages/phoenix-event-display/src/managers/ui-manager/phoenix-menu/phoenix-menu-ui.ts?rgh-link-date=2022-07-13T18%3A25%3A10Z#L224

    		const collectionNode = typeFolder.addChild(
     		  collectionName,
      >>		  (value: boolean) => {
      >>		    const collectionObject = this.sceneManager
       >>   		   .getObjectByName(SceneManager.EVENT_DATA_ID)
       >>   		   .getObjectByName(collectionName);
       >>		 this.sceneManager.objectVisibility(collectionObject, value);
       >>	    }
   			  );
Any idea why? 😅

9inpachi: Oh, that makes sense. That is a function which only executes on toggle.

Instead of spying on the `addChild`, you could check if the added child is in the tree by using `typeFolder.findInTree`.

DamianArado: cool!

packages/phoenix-event-display/src/tests/loaders/phoenix-loader.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    const collection = phoenixLoader.getCollection('Hits');
+
+    expect(collection).toEqual(collectionName);
+  });
+
+  it('get metadata associated to any event', () => {
+    // as of now returns an empty array of objects due to the event data we add in this test suite
+    const metadata = phoenixLoader.getEventMetadata();
+
+    expect(metadata).toEqual([]);
```
9inpachi, line 98: I get it now. It's empty because `buildData` accepts a single event not the file which contains multiple events.

So I recommend you pass `eventData['Event']` to `phoenixLoader.buildEventData` for this to work.

packages/phoenix-event-display/src/tests/loaders/phoenix-loader.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      'event number',
+      'run number',
+      'Hits',
+    ]);
+  });
+
+  it('should get the collection with the given collection name from the event data', () => {
+    const collectionName = eventData['Event']['Hits'];
+
+    const collection = phoenixLoader.getCollection('Hits');
```
9inpachi, line 89: Once you change the input to `buildEventData` like I suggested in the other comment. This will accept `hitsCollection` and return `eventData['Event']['Hits']['hitsCollection']`.

```ts
const expectedCollection = eventData['Event']['Hits']['hitsCollection'];

const collection = phoenixLoader.getCollection('hitsCollection');

expect(collection).toEqual(expectedCollection);
```

9inpachi (changes requested):
I added a comment about the metadata.

Just ping me whenever you are done. We should be able to merge this after these changes are done. :)

packages/phoenix-event-display/src/tests/loaders/phoenix-loader.test.ts
9inpachi: Or you can pass in an empty object `threeManager.init({})`.

DamianArado: If we use the real `ThreeManager` and initialize it, we know it calls the constructor for the `RendererManager` and this line throws up the error which we have been seeing since the start -

https://github.com/HSF/phoenix/blob/50d629e7dbf4432e3f1dd8b9a49466b53ecc0fc8/packages/phoenix-event-display/src/managers/three-manager/renderer-manager.ts#L20

```
Error creating WebGL context.
```

In these types of situations, we have to use a mocked ThreeManager (+ its functions) every time we need it in our tests.

9inpachi: The mock is fine. Looks okay now. But I think one last improvement could be to use the same object but just mock one method.

```ts
beforeEach(() => {
  phoenixLoader = new PhoenixLoader();

  const infoLogger = new InfoLogger();
  const threeManager = new ThreeManager(infoLogger);

  jest.spyOn(threeManager, 'addEventDataTypeGroup').mockReturnValue(new Group());

  phoenixLoader.buildEventData(
    eventData['Event'],
    threeManager,
    uiManager,
    infoLogger
  );
});
```

DamianArado: cool!

9inpachi (approved):

PR #477, opened 2022-07-11
Set transparency to false by default for LHCb

sponce:
This actually fixes most of the problems of objects not shown inside a transparent volume.

@9inpachi @EdwardMoyse  Out of curiosity : is there really a reason to have this default transparency explicit, and be able to set it to true ? It kind of breaks the attempts of three.js of rendering things properly by rendering non transparent objects first and transparent ones only after.

### DIFF
--- a/packages/phoenix-ng/projects/phoenix-app/src/app/sections/lhcb/lhcb.component.ts
+++ b/packages/phoenix-ng/projects/phoenix-app/src/app/sections/lhcb/lhcb.component.ts
@@ -93,7 +93,7 @@ export class LHCbComponent implements OnInit {
       undefined, // menuNodeName
       1, // scale
       true, // initiallyVisible
-      true // transparent
+      false // transparent
     );
 
     this.eventDisplay


9inpachi:
I understand but if we don't initially set it to true for the LHCb geometry then the opacity of some detector parts cannot be changed. You can just try running Phoenix on this branch locally and change the transparency of the detector. You will notice that the opacity doesn't apply to some parts. We introduced this option to make that work. And we added the option as a parameter because having transparency set to true initially caused problems in the ATLAS detector geometry.

I am not sure if it's a geometry issue but if you can fix the LHCb geometry so that the change in opacity works for all parts, then we won't need this option.

sponce:
Ok, now I see the original problem. Now, having a top level opacity handle that changes the opacity of the whole detector can hardly work properly anyway. As indeed it means that all objects are transparent and thus you will have troubles that they are not rendered properly, unless you manage to have your own sorting of objects if ever such a sorting is existing (I doubt it). So I would rather drop that top level changing of opacity.

But let's discuss that in the meeting this afternoon.

9inpachi:
Sorry, I was in a pairing session and missed today's meeting.

> So I would rather drop that top level changing of opacity.

The problem exists only in LHCb. If you are fine with it, we can remove the option. It shouldn't effect the other experiments.

sponce:
ok, I've moved the cleanup of this to a separate issue and I propose we merge this trivial MR in the meantime

PR #481, opened 2022-07-13
Merge meshes to improve performances

ahennequ:
* Merge meshes of the detector that share the same materials
* Merge Calo Cells
* Implement a tracks mesh that contains multiple instance of tracks (to reduce the number of draw calls) and implement features from https://github.com/spite/THREE.MeshLine

On the LHCb event viewer the framerate goes from 7fps to above 40fps.


ahennequ:
I've disabled tracks and calo optimisations while i'm working on the selection (it could take time). In this state, the PR already provides a significant speedup (from 7fps to about 25). I've also added a fix for the sliders not being moved until they are released and improved the render order of transparent meshes.

PR #486, opened 2022-07-15
Use full path for object names

sponce:
In case of 2 objects with identical name "Mirror" living in 2 different menus "Rich1" and "Rich2", it was so far impossible to identify them uniquely, as both names were just "Mirror".
With this change, the names are "Rich1 > Mirror" and "Rich2 > Mirror". Menu display has been adapted so that it stays the same

### DIFF
--- a/packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
+++ b/packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
@@ -300,11 +300,12 @@ export class ImportManager {
   private processGLTFSceneName(sceneName?: string, menuNodeName?: string) {
     if (sceneName) {
       const nodes = sceneName.split('_>_');
-      const name = nodes.pop();
       menuNodeName && nodes.unshift(menuNodeName);
-      const nodeName = nodes.join(' > ');
+      const fullNodeName = nodes.join(' > ');
+      nodes.pop();
+      const menuName = nodes.join(' > ');
 
-      return { name, menuNodeName: nodeName };
+      return { name: fullNodeName, menuNodeName: menuName };
     }
   }
 
--- a/packages/phoenix-event-display/src/managers/ui-manager/phoenix-menu/phoenix-menu-ui.ts
+++ b/packages/phoenix-event-display/src/managers/ui-manager/phoenix-menu/phoenix-menu-ui.ts
@@ -125,9 +125,20 @@ export class PhoenixMenuUI implements PhoenixUI<PhoenixMenuNode> {
       parentNode = this.geomFolder.findInTreeOrCreate(menuSubfolder);
     }
 
-    const objFolder = parentNode.addChild(name, (value: boolean) => {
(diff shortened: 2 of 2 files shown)


9inpachi:
I am not sure if there is a need to have unique names for the object? Is it something outside of the use cases Phoenix currently has?

If there is, I think a better way would be to assign the `fullNodeName` as the name of the object (as it's currently done) and specify the actual name in `userData.name` so there isn't much need to process the name in `phoenix-menu-ui.ts`. WDYT?

Either way, the changes will have to be done for `dat-gui-ui.ts` as well.

sponce:
I believe there is a need for unique names, simply to be able to retrieve them. How do I change something on the "Mirrors" of our Rich1 and not on the ones of Rich2 if I cannot name them ? This is actually what triggered this MR (although in our VP).

I'm not sure I understand the better way. Is what you call "actual name" the last part of what I would call the actual name, that is the full name ? You would basically store that part twice ? May be misleading to have 2 names with different contents. And anyway the only real name has to be the full one I believe.

9inpachi:
> I'm not sure I understand the better way. Is what you call "actual name" the last part of what I would call the actual name, that is the full name ?

Yes.

Okay, if you think it's fine this way, then please make the changes for `dat-gui-ui.ts` as well. And also other loaders maybe using this format. Would be great if you could change those as well.

sponce:
Sure will do. Although only on Monday as I took the next days off

sponce:
I'm back to this, and looking at `dat-gui-ui.ts`, I do not see what I would have to fix. There is nothing depending on the geometry names there, unless I misunderstand the code (which is fairly possible). Could you point me more specifically to what will break ?

9inpachi:
This is the reference: packages/phoenix-event-display/src/managers/ui-manager/dat-gui-ui.ts:128

sponce:
Where and how is that one used ? How can I test it ? It looks to me like this is another implementation of a menu, which has no hierarchy, and is thus flat. If this is the case, you probably want to keep the whole name actually.

sponce:
shot-2022-07-26_17-52-37
Does not look bad to me

9inpachi:
That looks good.

Just to check, I tried increasing the path length where we have 3 to 4 parents and it seems it starts overlapping with the item below.

image

So I would recommend just displaying the name (without path). For the UI to not break in case some use case later has this kind of requirement.

sponce:
Ok, tried it and... failed : 
```
ERROR Error: Uncaught (in promise): Error: You already have a folder in this GUI by the name "Modules"
```
So I fear we NEED unique names there.
I can of course still change it and let people realize by themselves that they should have unique names, but it's not so nice.
I could also cut the names and add '... ' at the beginning when cut. Not necessarily sufficient though.

9inpachi:
All right, thanks a lot for testing that.

I guess, we don't have much of a choice in that case.

9inpachi (approved):

PR #487, opened 2022-07-18
Added button to close/open the VP

sponce:
relies on and includes #486


packages/phoenix-ng/projects/phoenix-app/src/app/sections/lhcb/lhcb.component.html
```diff
@@ -22,6 +22,7 @@
     [eventDataImportOptions]="eventDataImportOptions"
   ></app-io-options>
   <app-share-link></app-share-link>
+  <app-vp-toggle></app-vp-toggle>
```
9inpachi, line 25: I would recommend to create a separate directory under the `lhcb` directory for this component. Otherwise the `lhcb` component and the `vp-toggle` component both exist in the same directory (8 files).

sponce: What's the problem in having the 2 components in the same directory ? 8 files looks like a small number, and creating one directory per 4 files a kind of overkill, no ? Are there other implications ?

9inpachi: There are no technical implication but going this route generally leads to inconsistent code base. It's a standard in Angular to have separate folders for each component. If you use Angular CLI to generate the component, it will automatically put the component in a folder.

I haven't seen any Angular code base that has multiple components in the same directory. It just makes it hard to navigate through the code.

https://angular.io/guide/styleguide#locate

9inpachi: :D Yeah. It's not a problem when it happens once or thrice. But when it becomes a practice at some point later in time because of something that exists in the code base.

sponce: Ok, code moved to a subdirectory. This should thus be ready to merge, except it's on top of #486 which is still under discussion

9inpachi (approved):

PR #490, opened 2022-07-19
Add manual inputs for range cuts

EdwardMoyse:
Within ATLAS we have had issues where the jets or cells aren't visible because the upper bound range is too low. We can change this in the code, but that's not very helpful if we have a special event. Also, it's quite fiddly to do cuts with the slider sometimes.

So this is a WIP to add manual entries. Currently it looks like (you can see I haven't worried about styling yet).

image

Currently, this isn't really working anyway - entered values don't move the slider or update the display, so the link isn't working.

I tried moving to use ` <input type="number" [(ngModel)]="value">` in `phoenix-menu-item.component.html, but then I get:
```
Error: projects/phoenix-ui-components/lib/components/phoenix-menu/phoenix-menu-item/phoenix-menu-item.component.html:126:32 - error TS2339: Property 'value' does not exist on type 'PhoenixMenuItemComponent'.

126                   [(ngModel)]="value">
```

I think I need help @9inpachi ! ;-)

### DIFF
--- a/packages/phoenix-ng/projects/phoenix-ui-components/lib/components/phoenix-menu/phoenix-menu-item/phoenix-menu-item.component.html
+++ b/packages/phoenix-ng/projects/phoenix-ui-components/lib/components/phoenix-menu/phoenix-menu-item/phoenix-menu-item.component.html
@@ -108,46 +108,38 @@
                 "
               />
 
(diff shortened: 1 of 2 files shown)


9inpachi:
@EdwardMoyse Works now.

PR #491, opened 2022-07-26
Use Jest for unit testing in phoenix-ng - I

DamianArado:
As we discussed currently, most of the tests fail due to this:

```
    ../phoenix-event-display/src/managers/three-manager/controls-manager.ts:307:21 - error TS2339: Property 'Tween' does not exist on type '_Group'.  

    307           new TWEEN.Tween(camera.position)
                            ~~~~~
    ../phoenix-event-display/src/managers/three-manager/controls-manager.ts:417:21 - error TS2339: Property 'Tween' does not exist on type '_Group'.  

    417         ? new TWEEN.Tween(camera)
                            ~~~~~
    ../phoenix-event-display/src/managers/three-manager/controls-manager.ts:418:21 - error TS2339: Property 'Tween' does not exist on type '_Group'.  

    418         : new TWEEN.Tween(camera.position);
```

After this gets fixed, we can improve other test cases and fix other problems here.

Was not able to run the tests keeping jest config in root as it gave a lot of "module not found errors" and also some weirdos like: 

```
TypeError: pathsToModuleNameMapper is not a function
```
or
```
import { pathsToModuleNameMapper } from 'ts-jest';
^^^^^^

SyntaxError: Cannot use import statement outside a module
```
or
```
$ jest --no-cache --config ../../jest.config.js app.component
● Validation Error:

  Preset jest-preset-angular not found.  <= although it's there!
```

@9inpachi, I liked what you suggested but I think we should use different config files each for `phoenix-event-display` and `phoenix-ng` 
due to the Jest config options specific to Angular setting: `preset`, `globalSetup`, `moduleNameMapper`, `setupFilesAfterEnv`, etc but maybe I'm wrong regarding this and those errors can be fixed anyways, we'll check.


9inpachi:
I see changes of some recent commits to master here. Could you try merging the target branch (jest-phoenix-event-display) with master? It seems both the jest branches have diverged a bit.

I noticed you already did this at times, but just for emphasis that keep the main jest branch up-to-date with master as often as possible. Otherwise, we can end up in a conflict hell.

DamianArado:
> Could you try merging the target branch (jest-phoenix-event-display) with master? It seems both the jest branches have diverged a bit.

There are merge conflicts here probably due to something that I copy-pasted. Can't merge `jest-phoenix-event-display` with `master` but I can merge `jest-phoenix-app` with `master`.  

Should I revert back the commits on `jest-phoenix-event-display` and first do a git pull and then push the new changes? Otherwise, I can close this PR and create a new PR on the target branch `jest-phoenix-app` as it is already up to date and there won't be any conflicts.

9inpachi:
> create a new PR on the target branch jest-phoenix-app

That sounds better. Maybe try that. But it should also be okay to resolve the merge conflicts. If they are in tests, then you just need to make sure you keep your version.

9inpachi:
One question I had to just sync is that the test setup works fine if we use the same configuration as `phoenix-event-display`, right? The tweenjs error is gone with that?

DamianArado:
> But it should also be okay to resolve the merge conflicts. 

In the case of dependencies, it would be somewhat complex. We will maintain these target branches and make sure they are in sync with master and they don't interfere with each other as well. So, I'll close this and revert back the changes here as it was before the commits on this PR.

> One question I had to just sync is that the test setup works fine if we use the same configuration as phoenix-event-display, right? The tweenjs error is gone with that? 

Nope. There are some angular-specific jest config options that we need to enable but the tweenjs error is only what keeps coming while running those `phoenix-ng` tests. 

I tried adding this
```
      tsconfig: {
        rootDir: null,
        allowJs: true,
      },
```
but TS also complains about it in the IDE.

Screenshot (240) 

Tried installing `@types/tweenjs`, added a declaration file to suppress this but the error still showed up on running the tests.

9inpachi:
> Tried installing @types/tweenjs, added a declaration file to suppress this but the error still showed up on running the tests.

Make sure you have the types removed. Like I mentioned earlier, the version that we are using is old and it doesn't have types at all.

> Nope. There are some angular-specific jest config options that we need to enable but the tweenjs error is only what keeps coming while running those phoenix-ng tests.

It's weird. If we use the same configuration options as `phoenix-event-display` the error should go away. I will try to find some time to find a fix. Just keep trying things in the mean time. It is most probably related to combining the Angular specific options and the Jest configuration of `phoenix-event-display` in the right way.

DamianArado:
> Make sure you have the types removed. 

Sure.

>  It's weird. If we use the same configuration options as `phoenix-event-display` the error should go away. I will try to find some time to find a fix. Just keep trying things in the mean time. It is most probably related to combining the Angular specific options and the Jest configuration of `phoenix-event-display` in the right way.

Sure, I will keep trying to fix that in the meantime.

DamianArado:
Please add the base ref `jest-phoenix-app` in `HSF/phoenix` so that I can create a PR on that branch.

PR #495, opened 2022-08-23
Add e2e tests for Phoenix App

DamianArado:
To check, inside `packages/phoenix-ng` type: 
- `yarn cy:open` for interactive browser-based output of the e2e tests.
- `yarn cy:run` for spinning up an electron-based headless browser for CLI output (ref. ss. attached below).

Screenshot 2022-08-23 165407

If something needs to be added/changed, please let me know.


9inpachi:
Hi @DamianArado,

I have not yet taken a detailed look at the tests, but the tests I skimmed through seemed to test only if some elements exist? That is okay but it doesn't test the interactions that have to be done like hiding part of the detector, using the UI menu to change the theme. Of course, that doesn't have to be done for all experiments separately. Testing these interactions and behaviors on a single (and relatively simple) experiment should be enough. Something similar to the playground.

We may be able to accept this as a start but I think the e2e tests should be more thorough to be useful.

You can also make use of [Cypress snapshots plugin](https://github.com/meinaart/cypress-plugin-snapshots) to take image snapshots of expected changes.

9inpachi (changes requested):
Besides what I have already mentioned, I have one following comment.

- The cypress setup should exist in `packages/phoenix-ng/projects/phoenix-app` instead of `packages/phoenix-ng`.

packages/phoenix-ng/cypress/e2e/atlas.cy.ts
```diff
@@ -0,0 +1,78 @@
+describe('ATLAS', () => {
+  it('should be up and running', () => {
+    cy.visit('/atlas');
```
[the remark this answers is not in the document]
DamianArado: There are 2 options here:

- adding it inside a `beforeEach` which can be taxing on the amount of time it takes loading the atlas experiment before each `it`.
- take it out of `it` which gives an error: `Cannot call cy.visit() outside a running test`.

My motive here is to make sure it loads this page before we test the things.

packages/phoenix-ng/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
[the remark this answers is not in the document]
DamianArado: Sure, I will if we don't need to test any JSON-based input. 🤔

packages/phoenix-ng/cypress/e2e/index.cy.ts
```diff
@@ -0,0 +1,130 @@
...
+        .contains(
+          'Visualisation for TrackML. Shows how to write a custom event loader'
+        )
+        .should('exist');
+      cy.get('.btn-primary').eq(5).contains('Show').should('exist');
+    });
+
+    it('should navigate to the correct page', () => {
+      cy.get('.btn-primary').eq(0).click();
+      cy.url().should('eq', 'http://localhost:4200/#/playground');
```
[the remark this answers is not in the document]
DamianArado: > unnecessarily checking if elements exist

Right. Instead of checking for elements in DOM, I should make sure to check whether they are visible or not.

> what e2e tests actually are.

I was not sure earlier up to what extent I should test the workflow. But, I'm adding more scenarios now. 👍

9inpachi:
There is also a beforeAll.

packages/phoenix-ng/cypress/e2e/atlas.cy.ts
```diff
@@ -0,0 +1,78 @@
...
+  });
+
+  it('should display the ATLAS detector properly', () => {
```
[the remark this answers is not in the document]
DamianArado: > Use [Cypress snapshots plugin](https://github.com/meinaart/cypress-plugin-snapshots) instead here.

Sure, I'll use it instead of checking for DOM elements.

packages/phoenix-ng/cypress/e2e/atlas.cy.ts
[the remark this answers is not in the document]
DamianArado: It'll be better to put it inside a `before` hook.

DamianArado:
> setup should exist in `packages/phoenix-ng/projects/phoenix-app` instead of `packages/phoenix-ng`.

It needs a package file in the root directory as we are not using the `angular.json` config here. 

*Make sure that you have already run [npm init](https://docs.npmjs.com/cli/init) or have a node_modules folder or package.json file in the root of your project to ensure cypress is installed in the correct directory.* ([Ref.](https://docs.cypress.io/guides/getting-started/installing-cypress#npm-install))

9inpachi:
> It needs a package file in the root directory as we are not using the angular.json config here.

Have you tried it out. From the reference, it says that the `package.json` should be in the _root_ of the project. The subdirectories shouldn't matter as long as it finds `package.json` somewhere in the parent directories. I may be wrong here but using Cypress in a monorepo is pretty common.

DamianArado:
> I may be wrong here but using Cypress in a monorepo is pretty common.

You're right. 👍 Cypress config file stays as it is where the package file is but we can move the setup inside `phoenix-app`.

DamianArado:
I hope the setup looks alright now. I noticed that browser-based testing gave flaky results (so I won't suggest trusting it as we all have different local environments) but the CLI output was consistent.

Still, if something needs to change, please let me know.

9inpachi:
Hi Humayun,

Thanks for the changes so far.

I have the following comments.

- We don't need this many screenshots.
  - What I suggested was to have only a single screenshot that covered each experiment.
  - Unless something changes on the 3D event display, you don't need to take a screenshot. Just check if the expected elements exist (not all but only one to three elements).
  - You need to take a screenshot when something changes on the event display because it's rendered on a canvas and we can't really assert anything otherwise. Common scenarios that may need screenshots would be:
    - Hiding detector/event.
    - Zooming in/out.
    - Clipping geometry.
    - etc.
- Please avoid using `cy.wait` with a hard-coded number. It significantly slows down the tests.
  - Instead, use `cy.get().should('exist')` or some other `should` assertion to make sure that the expected state is reached.
- Please avoid using `@ts-ignore`. There should be a way to globally introduce a type that may not be provided by the library.
- Avoid using common classes like `.btn-primary` to access an element. The tests will immediately break if someone introduces a new button.
  - Either find a more suitable selector that will not change or add the `data-testid` attribute to elements so they can be queried in tests but do so only if a suitable selector can't be found.
- Please write scenarios that test behavior instead of just visual/static changes. For example:
  - `it should toggle geometry visibility from Phoenix menu`
  - `it should create a shareable link based on specified parameters` (this will test the share link dialog)
  - `it should switch between perspective and orthographic view modes`
  - etc.
  - These may be a bit too detailed or specific. You can also think about the most common workflows the user will go through and test those. Keep it minimum for now, may be 5 to 10 scenarios but keep it right. (i.e. focus on quality over quantity)
- Please use `*.test.ts` instead of `*.cy.ts` to keep the file names consistent.
- Please remove the extra options in the configuration. I see a lot boilerplate that we will not need to configure. Let them fall back to their defaults.
- Please use TypeScript for the following files if possible:
  - `packages/phoenix-ng/cypress/support/commands.js`
  - `packages/phoenix-ng/cypress/support/e2e.js`
- Instead of naming the test cases something like "Visual test case for TrackML document", you can write:
  - `it should show the TrackML event display`

9inpachi:
> Can we generate the snapshots as well as test them on the CI? It may happen that the snapshots that I generated on my local machine be different from those generated in the CI environment.

It should work fine. CI environment are usually more robust and consistent compared to local environments. But we should focus on the locally working version for now and think about the CI later.

DamianArado:
> What I suggested was to have only a single screenshot that covered each experiment.

Yes, I did one for each experiment but added the tests for interactions and behaviors only in CMS experiment. 

> Unless something changes on the 3D event display, you don't need to take a screenshot. Just check if the expected elements exist (not all but only one to three elements)

This means we don't need to test if the Phoenix menu changes (visually) or not when we interact with it. ? We have to test only the elements whether they exist or not.

> You need to take a screenshot when something changes on the event display because it's rendered on a canvas and we can't really assert anything otherwise

I agree. 👍

> Instead, use cy.get().should('exist') or some other should assertion to make sure that the expected state is reached.

Sure, but not sure if the ATLAS/LHCb experiment loads that quickly. 😅 Let me try..

> Please avoid using @ts-ignore. There should be a way to globally introduce a type that may not be provided by the library

It gives an error in my IDE but works fine with tests: `Property 'toMatchImageSnapshot' does not exist on type 'Chainable<Document>'` which makes me think that somehow ts is not able to infer the types for this `toMatchImageSnapshot` function. I tried this https://stackoverflow.com/questions/56319017/how-to-fix-typeerror-cy-document-tomatchsnapshot-is-not-a-function and using a custom command but it didn't work. Let me try again.

> but do so only if a suitable selector can't be found

I agree that those selectors are brittle.

> write scenarios that test behavior instead of just visual/static changes
>  5 to 10 scenarios but keep it right. (i.e. focus on quality over quantity)

Sure, I'll add them to the CMS experiment tests.

> remove the extra options in the configuration

Sure. 

> use *.test.ts instead of *.cy.ts to keep the file names consistent

But wouldn't that mean they can get executed along with the Jest tests inside phoenix-ng? We can ignore this cypress directory completely in that case. 🤔

> it should show the TrackML event display

Okay, I'll do it this way. This description will be the name of the snapshots.

9inpachi:
> This means we don't need to test if the Phoenix menu changes (visually) or not when we interact with it. ? We have to test only the elements whether they exist or not.

Yes, that should be okay.

> Sure, but not sure if the ATLAS/LHCb experiment loads that quickly. 😅 Let me try..

There is a default timeout for these commands which you can increase in the configuration.

> But wouldn't that mean they can get executed along with the Jest tests inside phoenix-ng?

No, just ignore the Cypress tests directory in Jest configuration and vice versa.

DamianArado:
@9inpachi, I tried a lot of times but needed to wait in case of the CMS experiment because the detector didn't load before 2s. The rest of the points that you mentioned have been taken care of.  🙂

Please check it locally whenever you get time. And, if anything else needs to be updated, I'll do so ASAP.

packages/phoenix-ng/projects/phoenix-app/cypress/e2e/cms.test.ts
```diff
@@ -0,0 +1,171 @@
...
+
+      cy.contains('Save state').should('exist');
+      cy.contains('Load state').should('exist');
+
+      cy.get('.phoenix-menu-item-right .item-config-backdrop').click();
+    });
+
+    it('should display correct options under event data dropdown', () => {
+      cy.get('.phoenix-menu-item-right > button').eq(0).click();
+      cy.get('.phoenix-menu-item-right > button').eq(1).click();
```
9inpachi, line 23: Unfortunately, the order of these buttons could change depending on if the geometry or event data loads first.

Maybe you can use a selector like this:

```ts
cy.get("app-phoenix-menu").contains("Event Data").closest(".item-expand");
```

9inpachi: If you can, the best approach would be to add `data-testid`s to Phoenix menu items. But that will need some changing in the Phoenix menu code which could be complex.

packages/phoenix-ng/projects/phoenix-app/cypress/e2e/cms.test.ts
```diff
@@ -0,0 +1,171 @@
...
+
+      cy.contains('Tracks').should('exist');
```
9inpachi, line 25: You can scope this first: `cy.get("app-phoenix-menu").contains('Tracks').should('exist');`

This is true for the 4 selectors here. Because `Tracks` is a word that could appear somewhere else later if we add something.

packages/phoenix-ng/projects/phoenix-app/cypress/e2e/cms.test.ts
9inpachi: You could also create a function:

```ts
const getPhoenixMenuItem = (itemName: string) => {
  return cy.get("app-phoenix-menu").contains(itemName);
}

getPhoenixMenuItem("Event Data").closest(".item-expand").click();
```

9inpachi: The above comments are for all selections in Phoenix menu.

packages/phoenix-ng/projects/phoenix-app/cypress/e2e/cms.test.ts
```diff
@@ -0,0 +1,171 @@
...
+      cy.document().toMatchImageSnapshot();
+
+      cy.get('app-phoenix-menu-item').eq(3).click();
+      cy.get('.phoenix-menu-item-right > button').eq(0).click();
+    });
+  });
+
+  describe('Phoenix Iconbar', () => {
+    it('should switch to use dark theme', () => {
+      cy.get('app-dark-theme').eq(0).click();
```
9inpachi, line 53: Scoping the selector would be better than using the index: `cy.get('app-ui-menu').find('app-dark-theme').`.

packages/phoenix-ng/projects/phoenix-app/cypress/e2e/cms.test.ts
```diff
@@ -0,0 +1,171 @@
...
+      cy.get('app-zoom-controls app-menu-toggle > button').eq(0).focus();
+      for (let n = 0; n < 20; ++n) {
+        cy.focused().click();
+      }
+      cy.document().toMatchImageSnapshot();
+    });
+
+    it('should do object clipping properly', () => {
+      cy.get('app-object-clipping').click();
+      cy.contains('Clipping').click();
```
9inpachi, line 70: **Try to scope all the selectors that use `contains` or if possible, avoid using `contains` as a selector.** It seems very prone to breaking tests later on.

packages/phoenix-ng/projects/phoenix-app/cypress/e2e/cms.test.ts
```diff
@@ -0,0 +1,171 @@
...
+    it('should zoom in', () => {
+      cy.get('app-zoom-controls app-menu-toggle > button').eq(1).focus();
+      for (let n = 0; n < 20; ++n) {
+        cy.focused().click();
+      }
+      cy.document().toMatchImageSnapshot();
+    });
+
+    it('should switch to orthographic view', () => {
+      cy.get('app-main-view-toggle').eq(0).click();
```
9inpachi, line 94: Again, scope the selector.

packages/phoenix-ng/projects/phoenix-app/cypress/e2e/index.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    cy.go('back');
+    cy.get('.card a').eq(4).click();
+    cy.url().should('eq', Cypress.config().baseUrl + 'cms');
+    cy.go('back');
+    cy.get('.card a').eq(5).click();
+    cy.url().should('eq', Cypress.config().baseUrl + 'trackml');
+    cy.go('back');
+  });
+
+  it('should point to correct links in footer', () => {
```
9inpachi, line 30: I can't classify which type of test this would be. I don't think we need it. I would recommend removing it.

packages/phoenix-ng/projects/phoenix-app/cypress/e2e/user-scenarios.test.ts
```diff
@@ -0,0 +1,76 @@
...
+        '<iframe src="' +
+          Cypress.config().baseUrl +
+          'trackml?embed=true&file=test&type=json&config=test2"></iframe>'
+      )
+      .should('be.visible');
+
+    cy.contains('Close').click();
+  });
+
+  it('should rotate geometry', () => {
```
9inpachi, line 61: Does this produce the same result across multiple runs? If it doesn't, I think it's fine to remove it. It's a continuous animation and would be hard to consistently get the same image snapshot.

9inpachi (changes requested):
Great job with the new updates. The tests look really really better now! And of course, more meaningful. Thanks for the changes.

I have added some more comments after a more thorough review.

9inpachi:
@DamianArado Is this ready for review or are you still working on it?

DamianArado:
Yes, ready ✅. Can you please tell me if I need to change something else or
is it okay for now?

Future goal is to add more tests for each specific experiment (specifically
for Phoenix menu if their arrangement will not change sooner and if that
would be preferred).

packages/phoenix-ng/projects/phoenix-app/cypress/e2e/cms.test.ts
```diff
@@ -0,0 +1,190 @@
...
+      cy.get('.phoenix-menu-item').contains('Jets').should('exist');
+      cy.get('.phoenix-menu-item').contains('CaloClusters').should('exist');
+      cy.get('.phoenix-menu-item').contains('MuonChambers').should('exist');
+
+      cy.get('[data-testid="dropdown"]').eq(1).click();
+    });
+
+    it('should hide event data', () => {
+      cy.get('app-phoenix-menu-item .mat-slide-toggle')
+        .eq(0)
```
9inpachi, line 42: Does this not work for reaching a phoenix menu item? https://github.com/HSF/phoenix/pull/495#discussion_r957673738

The way it is now is still prone to errors since we are looking for any `.mat-slide-toggle` and getting it's first index and then making sure we have the `Event Data` option. This could break because the position of `.mat-slide-toggle` may change.

Please let me know if I misunderstand something here.

DamianArado: The problem when using that is - `Event Data` toggle gets triggered as well when we use the toggle for `Detector`. So, I fixed that. Pushing the changes..now it should work as expected.

9inpachi: That doesn't happen on my side. I used the following standalone test to check:

```ts
it('should hide detector geometry', () => {
  cy.get('[data-testid="dropdown"]').click();

  getPhoenixMenuItem('Detector').closest('.mat-slide-toggle').click();

  cy.document().toMatchImageSnapshot();

  getPhoenixMenuItem('Detector').closest('.mat-slide-toggle').click();

  cy.get('.phoenix-menu-item-right > button').eq(0).click();
});
```

This does not toggle "Event Data". Can you please share a screenshot how the "Event Data" gets triggered on clicking the "Detector" menu item?

DamianArado: Now, I got it why that happened with this line just before the `hide detector geometry` test case - 
```
cy.get('app-phoenix-menu-item').eq(1).click();
```

And due to this, the `Event Data` toggle gets triggered.

DamianArado: Does it look okay now or I need to change something else?  😅

9inpachi: Is okay now.

Just note that the snapshots don't work on my Linux because of differences in rendering the canvas. There are subtle differences in how WebGL works on different operating systems. I am not sure if it will work on some other Windows machine either. I will have to check. But just as a heads up, we may need to have a docker setup to make things consistent.

9inpachi (approved):
Thanks a lot for your efforts on this, @DamianArado!

packages/phoenix-ng/projects/phoenix-app/cypress/e2e/cms.test.ts
DamianArado: > we may need to have a docker setup to make things consistent.

For snapshots, definitely a yes! They have been inconsistent as per my check as well.

9inpachi:
On your machine, it should work if you set a diff threshold of 0.1. It's
0.01 by default which causes diff errors for even the minute details.

https://github.com/meinaart/cypress-plugin-snapshots#make-changes-to-default-configuration

DamianArado:
> On your machine, it should work if you set a diff threshold of 0.1. It's
0.01 by default which causes diff errors for even the minute details.

Right, checked. ✅

PR #512, opened 2022-11-02
Fixing loading of the vertex position

kjvbrt:
When using Phoenix loader, the position of the vertex was not properly loaded.


EdwardMoyse:
Ouch! Thanks for spotting this!

9inpachi (approved):

PR #513, opened 2022-11-08
fix(event-display): allowing geometry name to be undefined

kjvbrt:


packages/phoenix-event-display/src/event-display.ts
```diff
@@ -359,7 +359,7 @@ export class EventDisplay {
    */
   public async loadGLTFGeometry(
     url: any,
-    name: string,
```
9inpachi, line 362: Have you tested how the behavior is without the name? The name appears in the Phoenix menu and can potentially make it hard to navigate through the menu if there is no name.

EdwardMoyse: Test

kjvbrt:
Yes, but it is equally hard if you have only one name:
image

I'm not sure what is the preferred behaviour. To have the `name` variable as backup if the file does not contain the scene name or to overwrite the scene name from the file.

packages/phoenix-event-display/src/event-display.ts
EdwardMoyse: @kjvbrt is trying to reply... but in short, if he does not make it undefined then the passed name overrides the names in the geometry

EdwardMoyse: I think we have to accept this.

PR #515, opened 2022-11-09
feat(event-display): adds side and length for the calo cells

kjvbrt:
Adds side and length for the calo cells/clusters

### DIFF
--- a/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
+++ b/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
@@ -527,7 +527,7 @@ export class PhoenixObjects {
     defaultCellLength: number = 30
   ) {
     const cellWidth = clusterParams.side ?? defaultCellWidth;
-    const cellLength = clusterParams.length ?? defaultCellLength;
+    let cellLength = clusterParams.length ?? defaultCellLength;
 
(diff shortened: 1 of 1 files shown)


packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
```diff
@@ -515,25 +515,38 @@ export class PhoenixObjects {
...
-    length: number,
-    width: number = 30,
-    clusterParams: any
+    clusterParams: any,
+    defaultCellWidth: number = 30,
+    defaultCellLength: number = 30
   ) {
-    if (length < width) {
-      length = width;
+    let cellWidth = clusterParams.side;
```
9inpachi, line 529: You can use [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) for both `cellWidth` and `cellLength`..

```ts
const cellWidth = clusterParams.side ?? defaultCellWidth;
```

9inpachi (changes requested):

packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
kjvbrt: Thanks for the tip

PR #521, opened 2022-11-24
Make Vertices support both pos and x/y/z definitions + updated doc

sponce:
It seems different experiments started with different conventions here. It was simple enough to support both, so I implemented it

### DIFF
--- a/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
+++ b/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
@@ -700,10 +700,15 @@ export class PhoenixObjects {
     });
     // object
     const sphere = new Mesh(geometry, material);
-    sphere.position.x = vertexParams.pos[0];
-    sphere.position.y = vertexParams.pos[1];
-    sphere.position.z = vertexParams.pos[2];
-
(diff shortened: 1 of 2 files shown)


packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
```diff
@@ -700,10 +700,15 @@ export class PhoenixObjects {
...
-    sphere.position.z = vertexParams.pos[2];
-
+    if ('pos' in vertexParams) {
+      sphere.position.x = vertexParams.pos[0];
+      sphere.position.y = vertexParams.pos[1];
+      sphere.position.z = vertexParams.pos[2];
+    } else {
+      sphere.position.x = vertexParams.x;
+      sphere.position.y = vertexParams.y;
+      sphere.position.z = vertexParams.y;
```
kjvbrt, line 710: Why z coordinate of a sphere equals y coordinate of the vertex?

sponce: oups... because I reintroduced the code that was removed some time ago, and reintroduced the bug that was there altogether ! Fixing it right now

9inpachi (approved):

PR #532, opened 2023-01-06
Adding a button to create pictures from the current scene

sponce:
The goal is to be able to create pictures of any size (user can enter it) for the current scene. Unlike the screenshot mode, the new tool directly creates the picture respecting the size and type of fitting given.
Current features are : 
  - width and heights can be entered by hand in the menu
  - type of fitting can be "Crop" or "Strech"
    + Crop mode will scale the current screen keeping the width/height ratio until the new size is fully covered and cut either sides or top and bottom
    + Strech mode will only scale width and height independently to the new values, distording the image
  - the info panel is added on the picture in the top left corner (not configurable)

There are a couple of minor issues right now :
  - svg logos in the info panel are not rendered if the svg image does not have width and height attributes. It seems to be a limitation of html2canvas, see https://github.com/niklasvh/html2canvas/issues/1897. I'm working on a fix, but in the meantime, I've added width and height to the lhcb svg logo
  - the menu appearing when clicking on the tool is not perfectly aligned. That's my lack of html skills. I think it can be merged like this but if someone can improve it, be my guest


EdwardMoyse:
I'm not sure why the CI is failing here - it seems strange. I'll try to reproduce locally.

sponce:
This is now ready to be merged from my side.

EdwardMoyse:
This looks useful @sponce - thanks!

I have a few comments I added as a review. In addition, did you see this:
image
? Not sure how important it is.

Then some comments, which perhaps we can address in a followup PR, if you want to get this in soon.

1. It might be nice to have an option to add/don't add the overlay?
2. I think it's getting a bit confusing having a screenshot mode, and this new button. Since they're related, we could possibly put them in a single screenshot menu item ... i.e. you click on screenshot and a menu pops up with two options, enter screenshot mode, directly save a picture. 
3. Also we could also improve the mouseover to be more verbose (i.e. Screenshot mode: phoenix will become fullscreen and the menu is suppressed; "Creates a picture from the current view: directly saves a png of arbitrary size from the current view")

sponce:
> In addition, some more general comments. Firstly, did you see this warning: image ? (Not sure how important it is.)

I did. I have no clue what it means though. I'll have a look. It comes for sure from html2canvas when creating the info panel part of the picture.

>     1. It might be nice to have an option to add/don't add the overlay?

Not sure we need an option, we should just display what is on the screen. But indeed, the overlay is not supported right now. To be worked on. Probably in another PR.

>     2. I think it's getting a bit confusing having a screenshot mode, and this new button. Since they're related, we could possibly put them in a single screenshot menu item ... i.e. you click on screenshot and a menu pops up with two options, enter screenshot mode, directly save a picture.

I must say I'm not a gui guy and I have a hard time finding out what will be easier to use for end users.
But I agree that the button bar is quite crowded and on top these 2 features are strongly related.
Shall we also do it in another MR ?

>     3. Also we could also improve the mouseover to be more verbose (i.e. Screenshot mode: phoenix will become fullscreen and the menu is suppressed; "Creates a picture from the current view: directly saves a png of arbitrary size from the current view")

That was trivial enough that I've already done it :-)

> Finally, I'm puzzled by the CI failures, but I could [reproduce them](https://github.com/HSF/phoenix/blob/main/guides/developers/test-setup.md#running-the-tests-locally) so I'm pretty sure they're real.

I get no failure locally, so I'm not sure how to find out what's going wrong. It has to do with the ci environment. Or maybe the version of jsroot used ?

sponce:
I've now sorted out the javascript warnings in the console.

sponce:
And now I'm able to reproduce the failure locally. I was not launching the right tests. Looking into it

EdwardMoyse:
> And now I'm able to reproduce the failure locally. I was not launching the right tests. Looking into it

I think it must be unrelated actually, since this is suffering too:
https://github.com/HSF/phoenix/pull/534

:-(

sponce:
I'm lost with the one. I can reproduce it with my modifications, but when I remove them, the problem is still there. Of course, starting from a clean clone without my modifications works. So I conclude that there are some caches somewhere which get polluted. And actually I got another proof of it : I've rebased on main and then removed my stuff (so I have main) and it's failing but with different errors (and massively). So clearly there is hysteresis in the setup, which makes it really hard to debug

EdwardMoyse:
Since the problem is unrelated, let’s just merge this one in.

PR #536, opened 2023-01-15
Adding support for IrregularCaloCells physics objects

nilotpal09:
Calorimeter cells with arbitrary polyhedron-type geometry. 
Each cell is represented by 8 (x,y,z) vertices connected using a convex hull from the [ConvexGeometry](https://threejs.org/docs/#examples/en/geometries/ConvexGeometry) class.

IrregularCaloCells is a list of objects with the following attributes :
   * `layer` - the Calorimeter layer
   * `vtx` - a flattened list of 8 vertex coordinates (24 floats)
   * `color` - an integer list of [R,G,B] values
   * `opacity` - value from 0 to 1

### DIFF
--- a/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
+++ b/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
@@ -1,6 +1,7 @@
 import {
   Vector3,
   Object3D,
+  Color,
   CatmullRomCurve3,
   TubeGeometry,
   MeshToonMaterial,
@@ -21,6 +22,7 @@ import {
   LineSegments,
   LineDashedMaterial,
 } from 'three';
+import { ConvexGeometry } from 'three/examples/jsm/geometries/ConvexGeometry.js';
 import { EVENT_DATA_TYPE_COLORS } from '../../helpers/constants';
 import { RKHelper } from '../../helpers/rk-helper';
 import { CoordinateHelper } from '../../helpers/coordinate-helper';
--- a/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
+++ b/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
@@ -782,4 +782,43 @@ export class PhoenixObjects {
 
     return object;
   }
+
+  /**
+   * Type for drawing irregular calorimeter cell comprising 8 xyz vertices in arbitrary geometry.
+   * @param layer Calorimeter layer
+   * @param vtx Flattened list of 8 vertex coordinates (24 floats)
+   * @param color [R,G,B] integer list
+   * @param opacity value from 0 to 1
+   * @returns the cell
+   */
+   public static getIrregularCaloCell(irrCells: {    
+    type: any,
+    layer: number,
+    vtx: any,
+    color: string,
+    opacity: any,
+   }): Object3D {
+
+    const verticesOfCube = [];
+    for (let i = 0; i < 24; i += 3) {
+      verticesOfCube.push( new Vector3(irrCells.vtx[i],irrCells.vtx[i+1],irrCells.vtx[i+2]));
+    }
+    const geometry = new ConvexGeometry(verticesOfCube);
+    const cell_color = new Color("rgb(" + irrCells.color[0].toString() + "," + irrCells.color[1].toString() + "," + irrCells.color[2].toString() + ")");
+
+    // material
+    const material = new MeshPhongMaterial({
+      color: cell_color,
+      transparent: true,
+      opacity: irrCells.opacity
+      // wireframe: true
+    });
+
+    // object
+    const cell = new Mesh(geometry, material);
+    cell.userData = Object.assign({});
+    cell.name = 'IrregularCaloCell';
+
+    return cell;
+  }
 }
--- a/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
+++ b/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
@@ -785,12 +785,13 @@ export class PhoenixObjects {
     return object;
   }
 
-  /**
+/**
    * Type for drawing irregular calorimeter cell comprising 8 xyz vertices in arbitrary geometry.
-   * @param layer Calorimeter layer
-   * @param vtx Flattened list of 8 vertex coordinates (24 floats)
-   * @param color [R,G,B] integer list
-   * @param opacity value from 0 to 1
+   * @param irrCells Properties of irregular calo cell.
+   * @param irrCells.layer Calorimeter layer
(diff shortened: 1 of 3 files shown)


nilotpal09:
tagging @etiennedreyer

packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
```diff
@@ -782,4 +784,43 @@ export class PhoenixObjects {
 
     return object;
   }
+
+  /**
+   * Type for drawing irregular calorimeter cell comprising 8 xyz vertices in arbitrary geometry.
+   * @param layer Calorimeter layer
```
9inpachi, line 790: Since the function parameter is an object, all `@param` docs can be written with the `irrCells.` prefix.

```ts
/**
   * Type for drawing irregular calorimeter cell comprising 8 xyz vertices in arbitrary geometry.
   * @param irrCells Properties of irregular calo cell.
   * @param irrCells.layer Calorimeter layer
   * @param irrCells.vtx Flattened list of 8 vertex coordinates (24 floats)
   * @param irrCells.color [R,G,B] integer list
   * @param irrCells.opacity value from 0 to 1
   * @returns the cell
   */
```

Reference: https://jsdoc.app/tags-param.html#parameters-with-properties

packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
```diff
@@ -782,4 +784,43 @@ export class PhoenixObjects {
...
+    }
+    const geometry = new ConvexGeometry(verticesOfCube);
+    const cell_color = new Color("rgb(" + irrCells.color[0].toString() + "," + irrCells.color[1].toString() + "," + irrCells.color[2].toString() + ")");
+
+    // material
+    const material = new MeshPhongMaterial({
+      color: cell_color,
+      transparent: true,
+      opacity: irrCells.opacity
+      // wireframe: true
```
9inpachi, line 816: If the material is not to be wireframed, I think we can remove this extra commented code.

packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
```diff
@@ -782,4 +784,43 @@ export class PhoenixObjects {
...
+    });
+
+    // object
+    const cell = new Mesh(geometry, material);
+    cell.userData = Object.assign({});
```
9inpachi, line 821: The `cell.userData` we set here is displayed when we are in selection mode and click on this physics object. If this is not set, only the object name without any properties will be shown. If this is not intended, then you can set the properties using `cell.userData = Object.assign({}, irrCells);`.

packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
```diff
@@ -782,4 +784,43 @@ export class PhoenixObjects {
...
+    cell.name = 'IrregularCaloCell';
+
```
9inpachi, line 823: We need to add another line here to set the `uuid` property in the parameter object to be able to link the object from the collections info panel to the 3D canvas. Just add this line of code and it should be good to go.

```ts
// Setting uuid for selection from collections info
irrCells.uuid = cell.uuid;
```

9inpachi (changes requested):
Thanks a lot for the PR! I am sure this new physics object works as intended but I added a few comments for some Phoenix specific things that may not have been clear.

EdwardMoyse:
Another comment - it seems like the linting check failed:
```
/home/runner/work/phoenix/phoenix/packages/phoenix-event-display/src/loaders/phoenix-loader.ts
  304:21  error  Replace `⏎········new·Cut('layer',·0,·10),⏎······` with `new·Cut('layer',·0,·10)`  prettier/prettier

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```
If you're using Visual Studio Code, you can add the [prettier plugin](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) and then 'Format Document' with this linting tool.

etiennedreyer:
Hi @9inpachi  thanks for reviewing and suggesting the changes. We implemented them.
@EdwardMoyse we ran the linting, and we also added an option for cut based on cell energy.
Let us know if there's any other issue/request.

EdwardMoyse:
No objections from my side, but I'll let @9inpachi comment too.

9inpachi (approved):
Looks good to me too. Thanks for making the changes. :)

packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
```diff
@@ -782,4 +784,55 @@ export class PhoenixObjects {
...
+      opacity: irrCells.opacity,
+    });
+
+    // object
+    const cell = new Mesh(geometry, material);
+    cell.userData = Object.assign({}, irrCells);
+    cell.name = 'IrregularCaloCell';
+
+    // Setting uuid for selection from collections info
+    irrCells.uuid = cell.uuid;
```
sponce, line 834: There is no uuid in irrCells ! So my ci (LHCb's deployment one) gives the following error : 
```
src/loaders/objects/phoenix-objects.ts(834,14): error TS2339: Property 'uuid' does not exist on type '{ type: any; layer: number; vtx: any; color: string; opacity: any; }'.
```
I do not get why we did not see it in the official repo ! Can you fix it ?

packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
```diff
@@ -782,4 +784,55 @@ export class PhoenixObjects {
...
+   * @param irrCells.color [R,G,B] integer list
+   * @param irrCells.opacity value from 0 to 1
+   * @returns the cell
+   */
+  public static getIrregularCaloCell(irrCells: {
+    type: any;
+    layer: number;
+    vtx: any;
+    color: string;
+    opacity: any;
```
etiennedreyer, line 802: Hi @sponce we looked together and are not fully sure how to address this. We checked that the browser runs for us (with and without IrregularCaloCells) before pushing, but somehow it's not working the LHCb ci. 
Should we just add the following line here?
`    uuid: string;`

sponce: In the meantime, we've identified what might be the reason for different behavior in LHCb : we somehow use threejs 148 rather than 147. So I suppose moving back to 147 would make it work for us.

On the other hand, the code in indeed weird, as you set a member that does not exist. So indeed, I would add `uuid:string` or alternatively drop the line setting the uuid if it's not used. But I suppose it is in your case (for lhcb, I did that stop gap fix yesterday and it worked)

Issue #339, opened 2021-07-19
Make it possible to disable cuts or make it so if cuts are at bounds, then they don't count


EdwardMoyse:
I worked out the problem with #336. It's very simple - the jets are failing the energy cut. Now, I could increase the maximum to something huge work around this, but I think a better solution would to have a checkbox to enable/disable cuts. 

Another option would be to make it so that if the cut == max, then it doesn't count - i.e. if lower cut == min, then there is no < cut, and it upper cut == max, then there is no > cut. But this would require changing the GUI to show the infinity symbol or something in these cases. So I think option 1 is cleaner.

<img width="1089" alt="Screenshot 2021-07-19 at 10 12 23" src="https://user-images.githubusercontent.com/6764617/126126717-5d42da6b-bc05-4790-a40e-2980b4ee3952.png">

9inpachi:
Interestingly, the "Random cuts" button works in a similar way.

EdwardMoyse:
Fixed with #537

PR #537, opened 2023-01-16
Add ability to have open-ended range cuts

EdwardMoyse:
This PR is intended to add functionality to the range slider so that we can disable either end of the range cut. This (will be) a fix for #339 

I also refactored phoenix-menu-ui.ts `addCollecton` as part of this, adding `addDrawOptions` and `addCutOptions`.

There are, however, some issues right now that I'm struggling with. The most important is that selecting the slider toggles the min cut (see the animation below).

Animation3

Since I'm now running out of time to work on this, I thought I would open it up to others to comment.

Edit: when I drag the slider, I hit a breakpoint I set in setEnableMin (`phoenix-menu-ui.ts:315`)


packages/phoenix-event-display/src/managers/ui-manager/phoenix-menu/phoenix-menu-ui.ts
```diff
@@ -240,6 +240,99 @@ export class PhoenixMenuUI implements PhoenixUI<PhoenixMenuNode> {
...
+        value: cut.minValue,
+        highValue: cut.maxValue,
+        enableMin: cut.minCutActive,
+        enableMax: cut.maxCutActive,
+        onChange: ({ value, highValue }) => {
+          cut.minValue = value;
+          cut.maxValue = highValue;
+          this.sceneManager.collectionFilter(collectionName, cuts);
+        },
+        setEnableMin: (checked: boolean) => {
```
EdwardMoyse, line 315: This gets called when I drag the slider.

packages/phoenix-event-display/src/managers/ui-manager/phoenix-menu/phoenix-menu-node.ts
```diff
@@ -163,6 +163,8 @@ export class PhoenixMenuNode {
         value: config?.['value'],
         highValue: config?.['highValue'],
       });
+      config.setEnableMin?.(config?.['enableMin']);
```
9inpachi, line 166: I think this may be what is causing the behavior. You can try removing this just to see if it works. What may happen is that the UI in Angular rerenders and reconstructs the node in the tree and calls these functions.

packages/phoenix-ng/projects/phoenix-ui-components/lib/components/phoenix-menu/config/config-slider/config-slider.component.ts
```diff
@@ -8,12 +9,36 @@ import { Component, Input, Output, EventEmitter } from '@angular/core';
...
+    // We should probably disable the min input too, and change text to be infinity symbol or something?
+  }
+
+  toggleMaxCut(change: MatCheckboxChange) {
+    const value = change.checked;
+    if (value) {
+      this.max = this.oldMax;
+    } else {
+      this.oldMax = this.max;
+      this.max = Number.MIN_SAFE_INTEGER;
```
9inpachi, line 40: This should be `Number.MAX_SAFE_INTEGER`?

I am not sure if this is the source of the behavior you mentioned.

9inpachi (changes requested):
I took a quick look but if the suggestions don't work let me know and I can try to dive deeper.

packages/phoenix-ng/projects/phoenix-ui-components/lib/components/phoenix-menu/config/config-slider/config-slider.component.ts
EdwardMoyse: I don't think so - I don't think this code is ever called (I will remove it).

packages/phoenix-event-display/src/managers/ui-manager/phoenix-menu/phoenix-menu-node.ts
EdwardMoyse: No, I don't think so: I commented it out and no change (and I'm not sure why it would only affect the min checkbox and not the max?)

I fear it's something I did wrong in the html (which is what I'm weakest on!) - i.e. the checkbox is somehow part of the slider (or vice versa). But I can't see it if so...

EdwardMoyse:
Hi @9inpachi I'm still stuck on this. Any bright ideas you might have would be much appreciated!

9inpachi:
Hi Ed, sorry I have been a bit busy. I will take a look if I get the time.

EdwardMoyse:
> Hi Ed, sorry I have been a bit busy. I will take a look if I get the time.

Hi @9inpachi - I'll have a look at this again this week, but if you have time too that would be excellent (because I think you will have a better chance than me!)

9inpachi:
Hi @EdwardMoyse, I looked at it and the bug is because we have the config wrapped in a label. And if a checkbox is wrapped in a label, clicking anywhere inside the label will trigger the checkbox. But it's a bit tricky in our case since we use the label for other configs.

But I was thinking that instead of having these checkboxes, isn't it sufficient to just check if the values are at min or max and then disable the cut if they are? That seems better to me in terms of UX.

9inpachi:
The bug is fixed now. The label is now linked manually only with the `color` and `select` config types.

EdwardMoyse:
> But I was thinking that instead of having these checkboxes, isn't it sufficient to just check if the values are at min or max and then disable the cut if they are? That seems better to me in terms of UX.

I'm not super happy with how it looks right now either, but unless we somehow indicate that this the cut is now disabled, I don't think this is good UX either. I would personally be very confused to see objects with pT above the max pT cut, unless it was grayed out or something. But very happy to iterate on this as I don't really like my solution.

We do need this functionality in ATLAS though, so for the moment I will merge.

And thank you VERY much @9inpachi - it would have taken me a long time to work that out!

PR #542, opened 2023-01-19
Adding support for edm4hep tracker hits

kjvbrt:
Adding support for the edm4hep (sim)trackerhits to be visualized as circular points.

Also: adds `uuid` to the irregular calo cells object

### DIFF
--- a/packages/phoenix-event-display/src/loaders/edm4hep-json-loader.ts
+++ b/packages/phoenix-event-display/src/loaders/edm4hep-json-loader.ts
@@ -345,6 +345,7 @@ export class Edm4hepJsonLoader extends PhoenixLoader {
           type: 'CircularPoint',
           pos: position,
           color: '#' + hitColor,
+          size: 2,
         };
         hits.push(hit);
       });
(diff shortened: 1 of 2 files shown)


rahmans1:
@kjvbrt Will this be merged soon or are there any more adjustments that need to be made?

EdwardMoyse:
Hi, sorry for the delay in reviewing - but I do have a few concerns here. The main issue for me is I think that we should have a minimal number of EDM objects we support, and we only add a new one if the data cannot be represented with an existing class. Or in other words, we should not add a new type `CircularPoint` if we basically just want to draw the existing `Point` type in a different way. The EDM and how we visualise the EDM should be kept separate. Similarly, if we want a new way to visualise points, rather than adding a new function to do this, let's add a new parameter to the existing function.

kjvbrt:
But the hits (edm objects) are already represented in three different ways, no? I mean, one can specify Point, Line and Box.
Is point considered edm object? Having parameter for point anyway splits the implementation, but introduces additional ifs into the hitsToPoints function.

EdwardMoyse:
Hi @kjvbrt - okay, let me try to explain a different way. A EDM4HEP hit is just like a hit from ATLAS or CMS. There is no difference in the data content. You personally prefer to draw it as a circle, but who is to say another EDM4HEP client would feel the same way? They might want to use the default (which is probably/possible more performant). A line is different because needs to have more information, and so needs to be a different type.

So what I think should happen is that instead of making a new CircularPoint, we should expand the GUI to allow drawing of hits as circular points so everyone can benefit, not just EDM4HEP clients. 

(If you aren't sure how to do this, this is of course fine - you can open an Enhancement ticket and one of us can implement it. )

PR #543, opened 2023-01-19
Fixes for make picture feature, specialy for retina displays

sponce:
Retina displays are special in that the number of CSS pixels and the number of hardware pixels is different. Due to this, special treatment is needed which was missing in the make-picture component

On the way, the icon for make-picture was fixed (was greyed out in non dark mode)


EdwardMoyse:
I'm afraid I've just introduced a conflict @sponce. Can you rebase or whatever to fix it? This should also make the CI pass...

PR #547, opened 2023-01-23
Upgrade to yarn modern

EdwardMoyse:
As per issue #545, here I try to update to yarn modern (i.e. yarn > 1.x)

Why this is a good idea: https://yarnpkg.com/getting-started/qa#why-should-you-upgrade-to-yarn-modern

N.B. This does not (yet?) include a migration to plug'n'play: https://yarnpkg.com/getting-started/migration#switching-to-plugnplay

I also removed the deprecated `defaultProject` from angular.json, which mean tweaking `yarn deploy`

Also updated to jsroot 7.3.0, which no longer requires using @9inpachi's special version...

I tried updating other packages, but this caused issues in the CI, so I will try again in a separate PR.

PR #548, opened 2023-01-23
Fix CI

EdwardMoyse:
I pulled a few fixes out of #547

Hopefully this fixes the CI.

cc @sponce since some of this affects code written by him. I don't think it very controversial, except perhaps my documentation of 'saveBlob' ;-)


sponce (approved):
All fine for me.
Thanks for having added the missing const. I still do not manage with javascript constnees I must admit. lines like :
```
    const ctx = outputCanvas.getContext('2d');
    ctx.fillStyle = bkgColor;
```
simply hurt my eyes :-D

PR #561, opened 2023-03-22
getCaloPosition fixed for #552

### LINKED ISSUE #552: ATLAS calo cells are no longer rendered in the correct place
See the screenshot below. They should be on a cylinder, but are on a sphere.
image

I assume this was caused by #519 (my fault for not checking carefully!)

EdwardMoyse:
Add a new parameter, cylindrical, which is true by default, just in case we ever have a non-cylindrical calorimeter.

Re-implement the radial truncation logic removed by #519.

Fixes #552

@kjvbrt I'm pretty sure this won't break anything for you, but perhaps you could quickly review it?


kjvbrt:
Yeah, looks fine to me.

PR #562, opened 2023-03-22
Just strip embed in experiment-link

### LINKED ISSUE #527: In embedded view, 'open full' link should open the same event
See for example:
https://atlas.cern/updates/briefing/search-heavy-W-bosons

when we click 'open full' it opens the default event, which is not what we want. It should be the same URL as the embedded event.

EdwardMoyse:
The idea then is we open the event with full controls in full phoenix.

Thanks to @9inpachi  for the code!

Fixes #527


EdwardMoyse:
Argh! I did NOT mean to merge.

PR #564, opened 2023-03-22
Add forceColourTheme configuration option

EdwardMoyse:
One request we had from ATLAS was to be able to force the colour scheme. Since (I think) we don't save the status of the configuration bar in the 'save state' json, I figured this was the only other work around.


EdwardMoyse:
My apologies to everyone for forcing this in but I need it ASAP and no one joined the weekly meeting today.

PR #565, opened 2023-03-22
Request to change experiment info font to monospace

EdwardMoyse:
Another request from ATLAS, to make the experiment info font monospaced


EdwardMoyse:
Also forcing this one in without proper review ... sorry! We can follow up if necessary.

PR #587, opened 2023-06-26
Add cartesian grid

Somya-Bansal159:
This PR contains a "Cartesian Grid" option in the UI Menu, that opens a mat-menu, that has some options to customize a cartesian grid, such as:
- Show Cartesian Grid
- Show Labels (i.e. coordinates on the axes)
- Show XY Planes
- Change the number of XY Planes
- Show YZ Planes
- Change the number of YZ Planes
- Show ZX Planes
- Change the number of ZX Planes
- Change sparsity of the gridlines


kjvbrt:
Hi Somya,

Looking at the menu bar, can you use different icon for the grid sub-menu?
For example this one:
https://fontawesome.com/icons/table-cells-large?f=classic&s=solid
or any more suitable :)

Is there a way to add counter next to number of planes?
For example, in the option: | <>-----------------   XYPlane | add counter like so: | <>-----------------   6 XYPlanes |

Can the Sparsity option be better aligned?

Finally, the grid planes have mixed colors, can those be unified that?

Best,
Juraj

kjvbrt:
One more thing, maybe it will be better to integrate the tool with the use of a floating window. So, there will be only one Eye icon the the menu bar, and there will be one item (link) in the mat-menu. Upon clicking this item in mat-menu floating window with all possible settings will appear.

kjvbrt:
Hi @Somya-Bansal159,

Nice progress :)

Would it be possible to have the number of planes updated while user manipulates the scale?

Also, the grid is not much useful without the numbers on the edges. Have you found the place where the scale for the detector is set?

Best,
Juraj

Somya-Bansal159:
Number of planes are being updated while sliding.

Labels option has been added to view the numbers on the grid.

Also, I found out the place where scaling was set to 1000, also I understood as to how this was working. The diameter of the inner detector is given as 2.1m in the [link](https://atlas.cern/Discover/Detector/Inner-Detector) you provided.

Now, using the 3d distance feature, the diameter comes to be about 2100, which is 1000 times 2.1. So, probably, to find the units in meter, we just need to divide the coordinates by 1000 :)

kjvbrt:
Excellent work :)

I believe `cm` will fit the best. @EdwardMoyse what do you think?
Also, I will have a look at the FCC detectors , to see if the scaling is the same in their case.

Looking at the position of the modal:
* Can it be shifted so it is not exactly over the grid, maybe to the lower left corner? Something like in this [printscreen](https://github.com/HSF/phoenix/assets/34742917/5ff0138f-c768-451b-860f-c263976e884a).
* Can you use the same default values as the last time?
* And also, can you style the close button the same as is in the import modal? (lower right, green background)

I like that the labels are separated from the grid :)
* Can you add little ticks ([sketch](https://github.com/HSF/phoenix/assets/34742917/fd3df600-1608-4ed5-a332-0979d40db5ae)) and elevate the numbers a bit from the axis itself?
* On the x-axis the numbers are parallel to the axis, but on the other two axes the numbers are perpendicular to the axis, can you change it so on all axes the numbers are parallel to the axis?

Best,
Juraj

kjvbrt:
I checked the FCC detectors:
Screenshot from 2023-07-03 14-13-08

The outer radius of the detector is 6000 mm.

Somya-Bansal159:
I have shifted the dialog in the bottom left.

Could you please elaborate what default values you are talking about?

I have styled the close button as in the import modal.

I added the ticks as you mentioned and rotated the numbers. Could you please confirm if the styling and rotation is alright?

sponce:
Hi, indeed very nice feature ! very nice work @Somya-Bansal159 !

@kjvbrt I'm not sure what you mean with "scaling", but when trying the new feature on LHCb, I have a couple of feature requests... sorry @Somya-Bansal159 

LHCb is a "special" detector in that it's not centered on the interaction point, so when using the new feature, we have something like this :
shot-2023-07-06_17-02-33
So we obviously need to be able to define the region where the grid lies, and not necessarily be symetric around 0 (specially in z in the LHCb case)

The second issue I see is about the way axis are labeled. See this "classical" view of the LHCb Velo detector : 
shot-2023-07-06_17-07-31
Obviously we need a mechanism to "revert" the labels when we rotate, so that they are always readable.

Somya-Bansal159:
I have set the previous default settings (i.e. sparsity = 2)

As for throwing 'cm' at the end, I am slightly confused. The diameter of the inner detector was scaled 1000 times its actual value in meters, so shouldn't we add "mm" in the end?

For the movement of the grid, I will add another PR, in which there will be two options (in the dialog box itself), one to shift the grid using mouse click (i.e. the origin of the grid will be shifted to the new clicked point), another by manually entering x, y, z values of the new desired origin. And as @kjvbrt said, grid can be shifted by multiple mouse clicks and cancelled by right click.
So, @sponce for the LHCb, to show the grid in the desired region, can it be added in the next PR, where the position of the grid can be changed?

In order to view the labels from every angle, I have updated the animation loop to always align the labels towards the main camera using "lookAt()" function. Now there is no need to rotate the numbers as per axes. They will automatically point towards the camera.

sponce:
>  for the LHCb, to show the grid in the desired region, can it be added in the next PR, where the position of the grid can be changed?

Yes, no problem. We are not in a hurry

kjvbrt:
Hi @Somya-Bansal159,

yes the current number on the scale are mm, can't they be divided by 10?

One very minor nitpick, can also X,Y,Z labels face the camera?

best,
Juraj

kjvbrt:
Can square brackets be used?
X [cm]

9inpachi:
Thanks a lot for your contribution. Next time, please make sure that the tests are passing before merging your change. This PR introduced errors in the `ViewOptionsComponent` tests.

@Somya-Bansal159 @EdwardMoyse

PR #588, opened 2023-06-27
Auto reload feature for cycle-event component

sponce:
First moved file loading code from the ui component event-data-explorer-dialog to the dedicated service file-loader, for easier reuse

Then added auto reload event features to the cycle-event component

The component now toggles between 3 modes :
  - no cycle
  - cycle as before over existing events
  - cycle and reload events when reaching the end
Last mode allows for easy live feed providing the source is a URL

On the way, 2 little bugs have been fixed :
  - start the cycle on first event and not a random number that may be past the last event and fail
  - not mess up with event numbers when changing the set of events, again could lead to be past last event and fail
  - immediately start the cycle when clicking, not waiting for a first interval


EdwardMoyse:
Hi @sponce - could you add some description about what this is supposed to do? (I think I can work it out, but it would be best to hear it directly from you!)

sponce:
Done : plain cut and paste of the commit comments.

PR #595, opened 2023-08-01
Resurrected (and improved) tooltip for cycle event tool

sponce:
This has been broken when the cycle+reload state was added


sponce:
The error in the test does not seem to be related to this MR

PR #599, opened 2023-08-09
feat(event-display): calo cell opacity

kjvbrt:
Adding possibility to specify opacity for a calorimeter cell.


EdwardMoyse:
I think I would again put how energy affects opacity as something users control (i.e. part of the GUI). The more popular EDM4HEP becomes, the less sense hardcoded energy scales will make. But, we can always revisit this later.

EdwardMoyse:
What I'm more worried about is why the CI didn't run ... :-/

EdwardMoyse:
Closing and reopening to run the CI

PR #600, opened 2023-08-10
feat(event-display): visualize mc particles as arrows

kjvbrt:
Show MC particles as arrows from their origin with size and direction proportional to their momentum.
Primarily hard scatter particles are shown, but the cut could be adjusted, which is based on MC particle generator status.
image


EdwardMoyse:
@kjvbrt - any news on this one?

PR #604, opened 2023-08-18
mergeGeometries instead mergeBufferGeometries

EdwardMoyse:
Remove deprecated threejs call


hegner:
@EdwardMoyse - I believe instead of rebasing onto main you force pushed your changes into main. And that makes the PR being empty and implicitly merged. Best you check your command line history

EdwardMoyse:
> @EdwardMoyse - I believe instead of rebasing onto main you force pushed your changes into main. And that makes the PR being empty and implicitly merged. Best you check your command line history

Yeah, I was coming to the same conclusion. I think I must have been working in the wrong window somehow, but this was on my office computer and I'm remote now... I'll check again next week.

PR #617, opened 2023-08-28
Add a functionality to highlight subdetectors

Somya-Bansal159:
This feature is similar to collections info, except that, it works for detector geometries.

The highlight feature is working as expected.

But as for the "move camera to object", I am a bit unsure, what it would seem like to zoom into the big detector geometries. As of now, I am zooming into the origin of the geometries, and then zooming out by (1000, 1000, 1000).

Can you suggest any suitable icon for it?

Also, in the "collections info", the highlight feature didn't seem to be working properly. I tried to fix that as well, by setting effectsManager antialiasing to false, whenever we need to highlight an object.


EdwardMoyse:
Thanks for this!

Is there any way to visualise the geometry as a tree? For example, TGC2 is part of the Muon Spectrometer, and so in the right hand menu is below that in the hierarchy. But in your new view, all the geometry is lumped together.

Also, in the event data collections we have a 'hide invisible objects' option, which is very helpful. It would be good to have something similar for the geometry.

> But as for the "move camera to object", I am a bit unsure, what it would seem like to zoom into the big detector geometries. As of now, I am zooming into the origin of the geometries, and then zooming out by (1000, 1000, 1000).

Maybe try this approach?
https://discourse.threejs.org/t/camera-zoom-to-fit-object/936/2
 
> Can you suggest any suitable icon for it?

Maybe https://fonts.google.com/icons?selected=Material%20Symbols%20Outlined%3Afeature_search%3AFILL%400%3Bwght%40400%3BGRAD%400%3Bopsz%4024
? 

> Also, in the "collections info", the highlight feature didn't seem to be working properly. I tried to fix that as well, by setting effectsManager antialiasing to false, whenever we need to highlight an object.

Can you explain a bit more what you mean by it wasn't working?

Somya-Bansal159:
> Is there any way to visualise the geometry as a tree? For example, TGC2 is part of the Muon Spectrometer, and so in the right hand menu is below that in the hierarchy. But in your new view, all the geometry is lumped together.

I can try to display a nested list type of view, and then allow the user to select a child in it. But, in the collections info as well, the last child of every hierarchy is listed. So, maybe we have to change it there as well.



> Can you explain a bit more what you mean by it wasn't working?

image

The highlight button is not highlighting (I am assuming that it means to draw an outline) the element. The eye button also zooms into the element, but when it calls the function "highlightObject", it seems to be doing nothing at all.

EdwardMoyse:
I think the highlighting works as expected?

Animation6

PR #621, opened 2023-09-20
Fix: restrict simultaneous grid translation, both by pointer and by values

Somya-Bansal159:
Disabled the grid translation, while the translation using mouse click is active.


EdwardMoyse:
Thank you @Somya-Bansal159 (and apologies for only merging this now!)

PR #622, opened 2023-09-27
Allowed to specify "no-cache" for file loading

sponce:
This is actually useful in case of automatic file reload when cycling on live events. Without it, new events are not loaded as the cache enters the game.

This included extending the file-loader API to allow any options to be passed when loading a file.

It allows the following syntax in json files uses by the loader component : 
```json
[
    {"name":"Live events.json",  "url":"https:\/\/.../liveEvents.json.zip", "nocache":"true"},
    {"name":"2023/FirstCollisions2023.json",  "url":"\/assets\/files\/lhcb\/FirstCollisions2023.json.zip"},
...
]
```


EdwardMoyse:
Sorry for the very long delay in accepting this @sponce

PR #629, opened 2024-01-09
Update Phoenix to three 0.160.0

EdwardMoyse:
As per the title, just a threejs update.

Tested locally, no problems spotted.


EdwardMoyse:
It seems like precision of some conversions changed. Since these are physically negligible, I just adopted the new values without further investigation.

PR #665, opened 2024-06-20
Fixed support for gltf and glb zip files

sponce:


EdwardMoyse:
@sponce - I think this can go in now?

sponce:
Yes, it's ready in principle

PR #668, opened 2024-07-17
Update of set-up-phoenix instructions for Angular 17+

DraTeots:
- Works for updated angular
- Uses the standalone controls recommended since Angular 17+

### DIFF
--- a/guides/developers/set-up-phoenix.md
+++ b/guides/developers/set-up-phoenix.md
@@ -10,6 +10,10 @@
 * [Setting up the event display](#setting-up-the-event-display)
   * [Create an experiment component](#create-an-experiment-component)
   * [Set up the route](#set-up-the-route)
+* [Resolving problems](#resolving-problems)
+  * [Angular version](#angular-version) 
+  * [npm error gyp ERR](#npm-error-gyp-ERR)
+  * [ng build errors](#ng-build-errors)
 
 ## Introduction
 
@@ -33,13 +37,13 @@ npm install -g @angular/cli
 Now, using the Angular CLI, create a new Angular app.
 
 ```sh
-ng new my-app --style scss --routing true
+ng new event-display-app --style scss --routing true --interactive false
 ```
 
 Make sure the newly created app works.
 
 ```sh
-cd my-app
+cd event-display-app
 ng serve --open
 ```
 
@@ -51,30 +55,10 @@ Now that you have an app set up. Install the `phoenix-ui-components` package to 
 npm install phoenix-ui-components
 ```
 
-#### Import `PhoenixUIModule`
-
-After installing the package, open the `src/app/app.module.ts` file in your app and add `BrowserAnimationsModule` and `PhoenixUIModule` in your module imports.
-
-```ts
-import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
-import { PhoenixUIModule } from 'phoenix-ui-components';
-
-@NgModule({
-  ...
-  imports: [
-    ...
-    BrowserAnimationsModule,
-    PhoenixUIModule,
-    ...
-  ],
-  ...
-})
-export class AppModule { }
-```
 
 #### Import required styles
 
-Since some Phoenix components use Bootstrap, you will need to add the the Bootstrap stylesheet in the `src/index.html` file of your app.
+Add the the Bootstrap stylesheet to the `src/index.html` file of your app.
 
 ```html
 <head>
@@ -84,17 +68,18 @@ Since some Phoenix components use Bootstrap, you will need to add the the Bootst
 </head>
 ```
 
-After this, go to the `src/styles.scss` file of your app and import the global Phoenix styles.
+#### Add Phoenix Theming
+
+
+Go to the `src/styles.scss` file of your app and import the global Phoenix styles.
 
 ```scss
 @import 'phoenix-ui-components/theming';
-
-...
 ```
 
 #### Set up assets
 
-And lastly, download [these assets](https://github.com/HSF/phoenix/tree/main/packages/phoenix-ng/projects/phoenix-ui-components/lib/assets) (icons and images used by the Phoenix UI components) and put them in the `src/assets` directory of your app.
+Download [these assets](https://github.com/HSF/phoenix/tree/main/packages/phoenix-ng/projects/phoenix-ui-components/lib/assets) (icons and images used by the Phoenix UI components) and put them in the `src/assets` directory of your app.
 
 With this, the app is ready and we can move onto setting up the event display in this app.
 
@@ -112,12 +97,12 @@ The next step would be to create an Angular component so that we can use the eve
 For this, navigate to the `src/app` directory of your app in the terminal and use the Angular CLI to generate a component.
 
 ```sh
-ng generate component test-experiment
+ng generate component main-display
 ```
 
-This will create a `test-experiment` folder with the component source files.
+This will create a `main-display` folder with the component source files.
 
-Now, open the `test-experiment.component.html` file and use the Phoenix UI components to set up the UI.
+Now, open the `main-display.component.html` file and use the Phoenix UI components to set up the UI.
 
 ```html
 <app-nav></app-nav>
@@ -129,7 +114,7 @@ Now, open the `test-experiment.component.html` file and use the Phoenix UI compo
 <div id="eventDisplay"></div>
 ```
 
-The `[rootNode]="phoenixMenuRoot"` specified here for the Phoenix menu will be a defined in `test-experiment.component.ts`.
+The `[rootNode]="phoenixMenuRoot"` specified here for the Phoenix menu will be a defined in `main-display.component.ts`.
 
 One can easily customize the app-ui-menu. There are 2 main ways :
   - just adding buttons at the end of it by inserting the corresponding components with the `app-ui-menu` declaration :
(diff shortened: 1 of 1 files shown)


guides/developers/set-up-phoenix.md
```diff
@@ -195,11 +184,16 @@ export class TestExperimentComponent implements OnInit {
...
-
 }
 ```
 
-### Set up the route
+(!) It is assumed that you use your data and detector geometry instead of links provided in 
+`eventFile` and `loadGLTFGeometry` above. For testing purposes, you may take geometry and events at: 
+
+- [geometry examples](https://github.com/HSF/phoenix/tree/main/packages/phoenix-ng/projects/phoenix-app/src/assets/geometry)
+- [event examples](https://github.com/HSF/phoenix/tree/main/packages/phoenix-ng/projects/phoenix-app/src/assets/files)
```
tmadlener, line 194: From the current paths that are in the `main-display.components.ts` file it's not entirely obvious what would be two example files here that lead to a consistent view in the end. If there is such a pair of paths / files, I think it would be nice to state them here, as it would then be more easily possible to check if things are working as expected.

tmadlener (commented):
Very nice thanks. I have effectively just one minor comment at the very end. Other than that I was able to follow these instructions easily and got to a working example. I have the following versions
```console
$ node --version
v20.15.1
$ npm --version
10.7.0
$ ng version

     _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/
    

Angular CLI: 17.3.8
Node: 20.15.1
Package Manager: npm 10.7.0
OS: linux x64

Angular: 17.3.12
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1703.8
@angular-devkit/build-angular   17.3.8
@angular-devkit/core            17.3.8
@angular-devkit/schematics      17.3.8
@angular/cli                    17.3.8
@schematics/angular             17.3.8
rxjs                            7.8.1
typescript                      5.4.5
zone.js                         0.14.8
```

One observation from me (which seems to be an issue in angular?), is that I get the following warning with `npm start`. Everything seems to be working as expected though.

```console
9:51:34 AM [vite] warning: 
/home/tmadlener/work/students/desy_summer_2024/event-display-app/.angular/cache/17.3.8/vite/deps/chunk-7CYI36RD.js
2600|        return import(
2601|          /* webpackIgnore: true */
2602|          module
   |          ^^^^^^
2603|        );
2604|      });
The above dynamic import cannot be analyzed by Vite.
See https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars#limitations for supported dynamic import formats. If this is intended to be left as-is, you can use the /* @vite-ignore */ comment inside the import() call to suppress this warning.

  Plugin: vite:import-analysis
  File: /home/tmadlener/work/students/desy_summer_2024/event-display-app/.angular/cache/17.3.8/vite/deps/chunk-7CYI36RD.js?v=7b1de038
9:51:34 AM [vite] warning: 
/home/tmadlener/work/students/desy_summer_2024/event-display-app/.angular/cache/17.3.8/vite/deps/chunk-L6DXAODK.js
409|          return import(
410|            /* webpackIgnore: true */
411|            "file://" + name
   |            ^^^^^^^^^^^^^^^^
412|          );
413|        }).finally(() => fs.unlinkSync(name));
The above dynamic import cannot be analyzed by Vite.
See https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars#limitations for supported dynamic import formats. If this is intended to be left as-is, you can use the /* @vite-ignore */ comment inside the import() call to suppress this warning.

  Plugin: vite:import-analysis
  File: /home/tmadlener/work/students/desy_summer_2024/event-display-app/.angular/cache/17.3.8/vite/deps/chunk-L6DXAODK.js?v=7b1de038
9:51:34 AM [vite] warning: 
/home/tmadlener/work/students/desy_summer_2024/event-display-app/.angular/cache/17.3.8/vite/deps/chunk-L6DXAODK.js
461|        return import(
462|          /* webpackIgnore: true */
463|          url
   |          ^^^
464|        );
465|      }
The above dynamic import cannot be analyzed by Vite.
See https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars#limitations for supported dynamic import formats. If this is intended to be left as-is, you can use the /* @vite-ignore */ comment inside the import() call to suppress this warning.
```
---- 
Another question related to npm vs yarn. phoenix itself seemed to have switched to yarn, and I have tried to also use yarn here instead of npm, i.e. the commands I ran, were
```bash
ng new event-display-app --style scss --routing true --interactive false
cd event-display-app
yarn set version berry # as recommended by main phoenix documentation
yarn add phoenix-ui-components
yarn add phoenix-event-display
ng generate component main-display
```
Then do all the necessary edits to the file, and then try to
```bash
yarn start
```
which will fail with the follwing:
```console
✘ [ERROR] Could not resolve "jszip"

    .yarn/__virtual__/phoenix-ui-components-virtual-7124f6bf67/5/.yarn/berry/cache/phoenix-ui-components-npm-2.16.0-baa536ee41-10c0.zip/node_modules/phoenix-ui-components/dist/fesm2022/phoenix-ui-components.mjs:40:18:
      40 │ import JSZip from 'jszip';
         ╵                   ~~~~~~~

  The Yarn Plug'n'Play manifest forbids importing "jszip" here because it's not listed as a
  dependency of this package:

    .pnp.cjs:10935:31:
      10935 │         "packageDependencies": [\
            ╵                                ~~

  You can mark the path "jszip" as external to exclude it from the bundle, which will remove this
  error and leave the unresolved path in the bundle.
```

I suppose the issue is actually one from phoenix, and the fix is to simply "switch off" the [Plug'n'Play](https://yarnpkg.com/features/pnp) by adding the following line to the `.yarnrc.yml` file
```yml
nodeLinker: node-modules
```
(and then re-doing `yarn install`)

Maybe this is worth mentioning as well in the *Resolving problems* section?

guides/developers/set-up-phoenix.md
DraTeots: That, I believe, @EdwardMoyse could help

DraTeots:
About this problem: 

```
....
The above dynamic import cannot be analyzed by Vite.
See https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars#limitations for supported dynamic import formats. If this is intended to be left as-is, you can use the /* @vite-ignore */ comment inside the import() call to suppress this warning.
```

TL;DR; Ignore it. 

This happens, because phoenix libraries depend on three.js and jszip which are so called "common js" libraries. Those libraries in turn depends on many other "common js" libraries. Those libraries sometimes has different ways to import things as more or less standardized importing system is still kind of new in js world (some libraries exist for decades). So Vite which is used under the hood of ng can't analyze the import and issue the warning. There are 2 ways to solve the warnings: use another libraries, which is considered to be not possible or to reconfigure angular. As far as I googled it is not obvious reconfiguration (i.e. not setting some flag to true) so for the sake of tutorial simplicity - ignore it. Or... please let me know if I am wrong here, and it is a matter of some flag.

DraTeots:
```
✘ [ERROR] Could not resolve "jszip"
```

I believe it is to be a phoenix bug as I see `jszip` as `phoenix-event-display` dependency, but it is not listed as `phoenix-ng-components` dependencies while components use jszip. I believe `yarn` here behave more responsible towards highlighting a possible error. Please, file a bug report with this. 

P.S. At the same time, personally I prefer npm as in 2024 it fails less than yarn =p.

guides/developers/set-up-phoenix.md
EdwardMoyse: Right, well someone rolling their own version of Phoenix is presumably going to have to add their own geometry and event data ... I wasn't really expecting that anyone would use what is in `src/assets`. Maybe it's best just to add a caveat beneath these lines saying "(Obviously the geometry and event data should match, so if you use the TrackML event data, you should use the matching TrackML geometry)."

tmadlener: Just for completeness: It's possible to get e.g. CMS easily loaded because it's one file. But IIRC example data only exists for ATLAS, but that is a much more "fragmented" detector. So in the end this would effectively mainly be the icing on the cake to make sure that the basic setup worked as expected. But I also understand that this might be more work than is worth the effort.

EdwardMoyse: ATLAS and CMS are very similar detectors, it's just that for the CMS demo it was fine just to have one detector with no ability to switch off/on sub detectors. But we already knew that ATLAS wanted to use Phoenix as a fully functional event display, and so this was not acceptable (and at the time Phoenix could only add separate menu items for separate geometry files). The other advantage of having lots of separate sub-detector geometries is it means we can use these to make the different versions of ATLAS (e.g. Run1, Run2, Run3, Run4) by just combining what remained from the previous run, and adding what was upgraded.

I guess I could pretty easily make a complete ATLAS detector though... and this might have the advantage of being substantially smaller in file size.... I will have a look.

EdwardMoyse:
Sorry for the delay! And BTW I have no particular preference for `yarn` vs `npm` @DraTeots. I'm happy to discuss changing, if you think this is worthwhile (I honestly do not have enough experience to make the call).

Issue #689, opened 2024-09-24
Reset cut is not updating view


EdwardMoyse:
See:
https://github.com/user-attachments/assets/beb7359b-fa5a-4326-beff-38dddd8d3d5e

But basically, if you modify the cuts, then click 'Reset', the cut.model is updated but not the view (you get in the situation where you are turning a checkbox on, but it is already enabled by being reset)

This is WAY outside my comfort zone, but I wonder if the issue is in code like `` that we have:

```
              <div *ngSwitchCase="'rangeSlider'" class="range-slider">
                <div class="range-slider-inputs d-flex justify-content-between">
                  <mat-checkbox
                    [checked]="config.enableMin" 
                    (change)="config.setEnableMin($event.checked)"
                  ></mat-checkbox>
```
whereas we should be using a two way binding, i.e. using `ngModel`

```
              <div *ngSwitchCase="'rangeSlider'" class="range-slider">
                <div class="range-slider-inputs d-flex justify-content-between">
                  <mat-checkbox
                    [ngModel]="config.enableMin" 
                    (change)="config.setEnableMin($event.checked)"
                  ></mat-checkbox>
```

Unfortunately this doesn't seem to work, so I guess it's not that easy.... 

Pinging @9inpachi and @emiliocortina, who also know WAY more about Angular than me....

EdwardMoyse:
I tried to debug this in:
https://github.com/HSF/phoenix/tree/main-debug-refreshing-views

My idea was that we could use the `ChangeDetectorRef` in `PhoenixMenuItemComponent` to force the view to be redrawn.

To do this, I added a new method:
```
  viewUpdate(): void {
    this.cdr.markForCheck();
    this.calculateConfigTop();
  }
```

I also changed the html to also call the new method `viewUpdate`:
```
                (click)="config.onClick(); viewUpdate()"
```

But this does not work as I would expect:
- when I click on gear next to the "Cut Options" e.g. for Tracks, then `calculateConfigTop` is called.
- if I e.g. disable the min phi cut, the cut is updated as expected.
- if I click "Reset Cuts" then the `onClick` function defined in `PhoenixMenuUI` is called, followed by the new `viewUpdate` method.
- However the view is not redrawn (the checkbox is still unclicked, and any modified cuts are not shown to have reset (but I confirmed that the cut itself has been reset).

My best guess is that this is because I'm invalidating the parent and not the child view. But I'm not sure what else to try...

EdwardMoyse:
[...]

PR #692, opened 2024-10-20
[#689] Reset cut is not updating view

emiliocortina:
https://github.com/HSF/phoenix/issues/689

### DIFF
--- a/packages/phoenix-event-display/src/managers/ui-manager/phoenix-menu/phoenix-menu-node.ts
+++ b/packages/phoenix-event-display/src/managers/ui-manager/phoenix-menu/phoenix-menu-node.ts
@@ -108,17 +108,15 @@ export class PhoenixMenuNode {
 
   /**
    * Add a config to the phoenix menu item.
-   * @param type Type of configuration.
-   * @param options Options for the config.
+   * @param config config to be displayed as a Phoenix Menu item.
    * @returns The current node.
    */
-  addConfig<T extends keyof PhoenixMenuConfigs>(
-    type: T,
-    options: Omit<PhoenixMenuConfigs[T], 'type'>,
+  addConfig(
+    config: PhoenixMenuConfigs[keyof PhoenixMenuConfigs],
   ): PhoenixMenuNode {
-    const configsLength = this.configs.push({ type, ...options });
+    this.configs.push(config);
     // Apply the values of config
-    this.applyConfigState(this.configs[configsLength - 1]);
+    this.applyConfigState(config);
     return this;
   }
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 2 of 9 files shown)


packages/phoenix-event-display/src/managers/ui-manager/phoenix-menu/phoenix-menu-node.ts
```diff
@@ -108,17 +108,13 @@ export class PhoenixMenuNode {
 
   /**
    * Add a config to the phoenix menu item.
-   * @param type Type of configuration.
-   * @param options Options for the config.
+   * @param config config to be displayed as a Phoenix Menu item.
    * @returns The current node.
    */
-  addConfig<T extends keyof PhoenixMenuConfigs>(
```
EdwardMoyse, line 115: So, the plan here is to avoid type inference and make it explicit in the caller? Is there an advantage to this i.e. was this the cause of the problem?

(Just for my understanding)

EdwardMoyse:
Hey Emilio - thank you so much for doing this! I confirm that the reset cut button now works as expected. However, I think the cuts no longer work (for some reason which is not obvious).

For example, go to https://hepsoftwarefoundation.org/phoenix/#/atlas and load the attached [config file](https://github.com/user-attachments/files/17522438/run336567_evt2327102923-12.json) - the screen should go blank. But when I run this locally, the jets (the blue cones) are still visible.

Less importantly, you can find out how to fix the CI failure 
[here](https://github.com/HSF/phoenix/blob/main/guides/developers/test-setup.md#running-the-tests-locally)

TL;DR: run `yarn lint:fix`

packages/phoenix-event-display/src/managers/ui-manager/phoenix-menu/phoenix-menu-node.ts
EdwardMoyse: ping @emiliocortina ?

EdwardMoyse:
Hmm. A bit more debugging, and I think the issue might be that cut dialog ownership is not correct - cuts work for the first Jet collection ("AntiKt4EMTopoJets_xAOD") but for the second ("AntiKt4LCTopoJets_xAOD") the `collectionName` in `collectionFilter` is still "AntiKt4EMTopoJets_xAOD", AND if both are visible, the cuts from the second apply to the first.

EdwardMoyse:
Yeah, this is definitely it. In `addObjectType` for the first collection there is no `configRangeSlider` on the cuts. However, once we get to the second collection there is. So in this case, the new configRangeSlider is never created here:
```typescript
 public getConfigRangeSlider(
    collectionFiltering: () => void,
  ): ConfigRangeSlider {
    if (this.configRangeSlider == undefined) {
      this.configRangeSlider = {
        type: 'rangeSlider',
```

emiliocortina:
This is great! I took a look a couple of days ago and I found it very weird that it was working for the tracks, but not for the jets

PR #711, opened 2025-02-13
fixes #567

### LINKED ISSUE #567: Order better the Collections info dropdown
When we open the Collections Info windows, the collections dropdown is currently unsorted. See below.

What would be ideal is to divide it into different types of collections. Or we add another dropdown to select the type, with this dropdown then only showing collections of the requested type.

If all of this is too hard (or as a stopgap) we should at least sort it alphabetically.

image

Pratz2005:
This PR fixes issue #567 - orders collections info dropdown better

Before:
image

After
image


Pratz2005:
Will look into the failing tests

EdwardMoyse:
> Will look into the failing tests

I think you just need to do:
```
yarn lint:fix
```

Issue #714, opened 2025-05-21
Inconsistent Cameras


AlphaAbdo:
**Description:**
There is a critical inconsistency in the zooming mechanism between the active **Perspective Camera** and the **Orthographic Camera**. This leads to the Orthographic mode being essentially **unusable** for standard tasks, such as visualizing the full detector.

### 🐞 Bug Behavior

When switching to the Orthographic Camera and attempting to zoom out (e.g., to frame the entire detector), the following occurs:

* The **Perspective Camera**'s position is updated as well, resulting in it moving too far back.
* This causes **scene elements to disappear**, including critical components like tracks.
* Because both cameras share the same position, and the Perspective camera interprets this position differently, zooming out breaks the scene rendering in both modes.

### Example

![Image](https://github.com/user-attachments/assets/d0077ae2-d188-4eee-a576-0be1c85fa063)

In the attached images, you can clearly see:

* After zooming out in Orthographic mode, **the back/mid scene layers vanishes**.
* Even before full disappearance, **tracks are missing** due to the extreme distance of the Perspective camera.

### 🔄 Expected Behavior

* Orthographic and Perspective cameras should manage zoom/position independently, or
* A mechanism should be added to decouple their zoom levels and field of view/focus behavior.

AlphaAbdo:
Hi! Thanks a lot for your interest in contributing to the Phoenix project ,  it’s always great to see new contributors getting involved!

Just a heads-up: I’ve already submitted a pull request that includes a fix for this issue along with several other related bugs (https://github.com/HSF/phoenix/pull/720) . However, this particular bug is relatively simple and could still be a great exercise for you to get familiar with the codebase.

If you're still up for it, I’d be happy to walk you through how the problem was tackled and point you to the relevant parts of the code. I’ve been working on the project actively for the past couple of months, so I can share some insights that might help you ramp up faster.

Let me know how you'd like to proceed !

Priansh-master:
Thanks so much for the quick response and for the warm welcome!

I'd definitely love to go through the issue as an exercise and get familiar with the codebase. It would be great if you could walk me through how you approached the fix and point me to the relevant parts of the code. That kind of guidance would really help me ramp up faster and start contributing meaningfully.

Looking forward to learning from you!

Issue #713, opened 2025-05-20
zoomTo in controls-manager.ts causes unintended behavior due to improper camera manipulation


AlphaAbdo:
**File:**
`main/packages/phoenix-event-display/src/managers/three-manager/controls-manager.ts`

**Function:**
`public zoomTo`

**Description:**
When performing a zoom artificially—either by clicking the UI zoom buttons (which inadvertently trigger the `zoomTo` function twice), or by using the keyboard shortcut `Shift + '+'` ; the camera behavior becomes problematic.

**Problem Details:**

* The root cause stems from handling the camera as a generic `Camera` rather than as a `PerspectiveCamera`.
* The current implementation only divides the camera’s position vector, which causes it to converge toward the origin `(0, 0, 0)` with repeated zooming.
* This approach fails to maintain the current orientation and leads to an unnatural zooming experience.
* As a result, it diverges from the behavior expected when zooming with mouse input (e.g., scroll wheel), which adjusts the camera’s position in a direction perpendicular to the view vector—preserving the camera's rotational state.

**Expected Behavior:**
The zoom should:

* Respect the `PerspectiveCamera` type,
* Move the camera in a direction perpendicular to the viewing direction (consistent with mouse-based zooming),
* Avoid arbitrary position scaling,
* Preserve the camera's rotation and orientation for a smooth user experience with Tween,

**Suggested Fix:**
Refactor `zoomTo` to:

* Safely cast the camera to `PerspectiveCamera`,
* Modify the position based on the current viewing direction (e.g., use camera.getWorldDirection or similar),
* Match the same logic used by the mouse wheel zoom listener.

EdwardMoyse:
Thank you very much for the details bug report(s). Since you seem to have a very good handle on where the issue is, would you be able to make some PRs to fix them?

PR #720, opened 2025-06-25
Controls-Manager expansion pack

AlphaAbdo:
This pull request is the result of group discussions during the June Phoenix meeting. As reported in Issues #714 and #713, several minor bugs—though not deemed critical—were persistent enough to warrant attention. I took the initiative to investigate and fix them. What initially appeared to be small issues turned out to be symptoms of deeper architectural flaws.

These fixes led to a partial rebranding of the `Control-Manager`, with the goal of treating the **Overlay** as an independent component. While it can still be bound to the main display, this behavior is now handled as a built-in feature rather than an assumption.

Several buttons were added to improve the user experience, and the overlay canvas management has been corrected ; allowing proper aspect ratio handling and preventing unintentional scaling behavior.

I will also include a visual comparison showcasing the new interface versus the previous one.

---

**Bugs fixed include (but are not limited to):**

* Camera desynchronization between main and overlay views.
* Overlay was always treated as a child scene.
* Random desync occurrences during usage.
* Image upscale on overlay window expansion, instead of resizing the canvas properly.
* Double input conflicts: both `mainControls` and `overlayControls` were listening to the same canvas, causing erratic behavior.
* Zoom button triggering the zoom functionality twice on a single click.

---
Old:
Screenshot 2025-06-25 144327

New:
Screenshot 2025-06-25 145747

### DIFF
--- a/packages/phoenix-event-display/src/managers/three-manager/controls-manager.ts
+++ b/packages/phoenix-event-display/src/managers/three-manager/controls-manager.ts
@@ -27,10 +27,12 @@ export class ControlsManager {
   /** The renderer manager instance for accessing main and overlay renderers. */
   private renderr: any;
 
+  /** Whether the overlay controls are linked to the main controls. */
   private _isOverlayLinked: boolean = false;
   /**
    * Constructor for setting up all the controls.
    * @param rendererManager The renderer manager to get the main renderer.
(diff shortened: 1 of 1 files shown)


sponce:
This PR resolves #713"
This PR resolves #714"

AlphaAbdo:
yarn test:ci

PR #729, opened 2025-10-28
Update angular to v20

EdwardMoyse:
Unfortunately this does not work yet. 

When I do `yarn start` I get 
```
node:worker_threads - Error: Module build failed: UnhandledSchemeError: Reading from "node:worker_threads" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
```
but opening here in case anyone has any ideas what I need to change?

cc @9inpachi @sponce @AlphaAbdo @DamianArado

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -8326,17 +8326,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"dunder-proto@npm:^1.0.1":
-  version: 1.0.1
-  resolution: "dunder-proto@npm:1.0.1"
-  dependencies:
-    call-bind-apply-helpers: ^1.0.1
-    es-errors: ^1.3.0
-    gopd: ^1.2.0
-  checksum: 149207e36f07bd4941921b0ca929e3a28f1da7bd6b6ff8ff7f4e2f2e460675af4576eeba359c635723dc189b64cdd4787e0255897d5b135ccc5d15cb8685fc90
-  languageName: node
-  linkType: hard
-
 "duplexer@npm:^0.1.1, duplexer@npm:~0.1.1":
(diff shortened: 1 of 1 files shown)


EdwardMoyse:
```
 YN0001: │ cypress-plugin-snapshots@git+ssh://git@github.com/meinaart/cypress-plugin-snapshots.git#commit=a8bd88380db56905f5df6fe27cb4eddb7b809b9e: Failed cloning the repository
  ➤ YN0001: │   Repository URL: ssh://git@github.com/meinaart/cypress-plugin-snapshots.git
  ➤ YN0001: │   Git@github.com Error: Permission denied (publickey).
  ➤ YN0001: │   Fatal Error: Could not read from remote repository.
  ➤ YN0001: │   Exit Code: 128
```
No idea why this works locally but not here... To be continued.

EdwardMoyse:
That's a good idea - I think there are forks of `cypress-plugin-snapshot`
which are on npm. Or indeed, probably https might work. What I don't
understand is what changed to break this? But the main thing is to get it
working again.

EdwardMoyse:
Locally at least, this still seems to be failing with:
```
node:worker_threads - Error: Module build failed: UnhandledSchemeError: Reading from "node:worker_threads" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
```

deveshbervar:
Thanks a lot for the invite, @EdwardMoyse!  
Really appreciate it — I’ll start from a clean branch and reopen the upgrade as a new MR.

PR #738, opened 2025-11-25
feat(screenshot): Add tiled rendering to support very large screenshots (#581)

### LINKED ISSUE #581: Problems with large screenshots

deveshbervar:
### Summary
This PR implements the remaining part of Issue #581 (Problems with large screenshots).

Large screenshots previously failed due to WebGL `MAX_RENDERBUFFER_SIZE` limits, causing zoomed-in or black output.  
The fix adds **tiled screenshot rendering**, allowing arbitrarily large screenshots by splitting the render into tiles and stitching them together.

### What this PR adds
- Detects max texture size from WebGL
- Splits the requested screenshot into multiple tiles
- Uses `camera.setViewOffset()` for each tile
- Renders tile-by-tile and draws them onto a large output canvas
- Restores renderer + camera state after completion
- Supports both Stretch and Crop modes (existing behavior preserved)

### Why this is needed
Browsers cannot render huge canvases in one pass.  
Tiling bypasses WebGL size limits and fixes the incorrect export output reported in #581.

### Issue Reference
Fixes #581

---

Thank you, and happy to iterate on feedback!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,7 +49,7 @@ ENV CI=1
 ENV CYPRESS_INSTALL_BINARY=0
 
 # Install dependencies (verbose so you can see if python errors occur)
-RUN yarn install --silent --verbose
+RUN yarn install --silent
 
 # Build the web app
 RUN yarn deploy:web


sponce:
Thanks a lot for that. Looks all good from a quick code inspection.
Now when trying it, the crop is perfect but the strech is somehow not working. See the 3 files attached : 
   - number 6 gives an idea of what I have on the screen (it's also cropped, but not too much)
   - number 4 is the cropped version to a "flat" image, and worked nicely
   - number 5 is the streched version to the same size and is not what is expected.
 
screencapture(4)
screencapture(5)
screencapture(6)

EdwardMoyse:
> @EdwardMoyse the ci is failing, but it looks like it's the machinery itself, not the PR. Am I correct ?

Not AFAIK. Other MRs are passing CI ... somehow it is related to this one.

EdwardMoyse:
Hi @deveshbervar - does `yarn test:ci` work for you locally? I'll try to look into this later, but I'm not sure why you need to update the dockerfile for this MR, when it's running fine in all the other CIs (though I have to admit I've not had time to look properly yet so perhaps I'm missing something obvious?)

deveshbervar:
Hi @EdwardMoyse, thanks for the review!
I reverted the callback typings back to MouseEvent and kept them initialized with no-op functions to avoid the TS “callable {}” error.
I also updated the Dockerfile by replacing python with python3, which fixes the CI package installation issue on Ubuntu 22.04.
Let me know if anything else needs adjustment!

PR #771, opened 2026-01-07
Fix memory leak in CycleEventsComponent by cleaning up interval

remo-lab:
**This PR fixes a memory leak in` CycleEventsComponent `where an active `setInterval` was not cleared when the component was destroyed.**
## Problem
`CycleEventsComponent` starts an interval to automatically cycle through events, but previously only implemented `OnInit. `When navigating away while cycling was active, the interval continued running in the background, which could lead to:
- Memory leaks
- Calls to `eventDisplay.loadEvent()` after component destruction
- Unnecessary background work and performance issues

## Fix
- Implemented the OnDestroy lifecycle hook
- Cleared the interval in ngOnDestroy() to ensure proper cleanup
- Added a defensive guard to prevent multiple intervals from running at the same time
- Added a safety check to avoid cycling when no events are available

## How to Test
- Run the app:
       `yarn start`
- Navigate to http://localhost:4200/lhcb
- Start event cycling using the cycle-events button
- Navigate to another route while cycling is active

**Expected result:
The interval is cleared on navigation, no background cycling continues, and no console errors occur.**

## Impact

- Prevents memory leaks during navigation
- Avoids runtime errors from destroyed components
- Improves performance and resource management
- No breaking changes

### DIFF
--- a/packages/phoenix-ng/projects/phoenix-ui-components/lib/components/ui-menu/cycle-events/cycle-events.component.ts
+++ b/packages/phoenix-ng/projects/phoenix-ui-components/lib/components/ui-menu/cycle-events/cycle-events.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input, type OnInit } from '@angular/core';
+import { Component, Input, type OnInit, type OnDestroy } from '@angular/core';
 import { EventDisplayService } from '../../../services/event-display.service';
 import { FileLoaderService } from '../../../services/file-loader.service';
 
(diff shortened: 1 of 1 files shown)


remo-lab:
Hello @EdwardMoyse ,
This PR fixes a memory leak in `CycleEventsComponent` by ensuring the interval is cleaned up on component destruction.
I would appreciate a review when you have time. Thank you.

PR #775, opened 2026-01-10
feat: make excluded columns configurable in CollectionsInfoOverlay

Sai-Adithya-M:
Description:

Validation
[x] Code fulfills the rules in CONTRIBUTING.md
[x]test:ci passed (locally)
What:
This PR refactors CollectionsInfoOverlayComponent to allow the list of excluded columns in the collections info table to be configurable via an @Input().

Why:
Previously, the list of excluded columns (['uuid', 'hits', 'isCut', 'labelText']) was hardcoded with a "FIXME" comment indicating it was an "ugly hack". Making this list configurable improves the reusability of the component and allows different contexts to exclude different columns as needed.

Changes
* Added @Input() excludeColumns: string[] to CollectionsInfoOverlayComponent
 (defaulting to the original exclusion list).
* Updated changeCollection method to filter columns based on this.excludeColumns.
* Added a unit test to verify that custom columns can be excluded.

### DIFF
--- a/packages/phoenix-event-display/src/loaders/jivexml-loader.ts
+++ b/packages/phoenix-event-display/src/loaders/jivexml-loader.ts
@@ -75,7 +75,6 @@ export class JiveXMLLoader extends PhoenixLoader {
    * @param data Event data in JiveXML data format.
    */
   public process(data: any) {
-    console.log('Processing JiveXML event data');
     this.data = data;
   }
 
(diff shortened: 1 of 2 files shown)


EdwardMoyse:
The linting error probably just need `yarn lint:fix`, but was the actual change implemented? Unless I'm missing something it seems to just be removing log / debug messages?

PR #783, opened 2026-01-21
fix(scene-manager): dispose GPU resources on event switch

goyalpalak18:
### **Description**

This PR addresses a critical memory leak that occurs when switching between events or clearing event data.

While the current implementation of `clearEventData()` successfully removes objects from the Three.js scene graph, it doesn't notify the WebGL renderer that the underlying GPU resources (geometries, materials, and textures) are no longer needed. In Three.js, these resources are not tracked by the JavaScript Garbage Collector and must be disposed of explicitly to free up VRAM.

Without this fix, every event switch causes a linear increase in GPU memory usage. For users browsing through large ATLAS/CMS events or running the display for extended periods, this inevitably leads to a WebGL context loss or a total browser tab crash.

### **Changes**

* **Refactored `clearEventData`**: Now ensures that every object is thoroughly "cleaned" before being detached from the scene.
* **Added `disposeObject3D` helper**:
* Recursively traverses the `Object3D` hierarchy.
* Calls `.dispose()` on all `BufferGeometry` instances.
* Handles `Material` disposal (including support for `Material[]` arrays).
* **Texture Cleanup**: Iterates through all standard material maps (map, alphaMap, normalMap, etc.) to ensure textures are removed from GPU memory.

### **How to verify**

1. Open the Phoenix application and monitor the browser's performance (e.g., Chrome DevTools -> Three dots -> More Tools -> **Performance Monitor**).
2. Watch the **GPU Memory** or **JS Heap** (indirectly affected).
3. Load a large event and switch events 20–30 times.
4. **Before fix**: GPU memory climbs steadily until the tab becomes sluggish or crashes.
5. **After fix**: GPU memory should return to a stable baseline shortly after each `clearEventData()` call.

### **Impact**

* **Stability**: Eliminates "Rats! WebGL hit a snag" errors during long analysis sessions.
* **Performance**: Improves frame rates for users on low-end hardware or integrated GPUs.
* **UX**: Allows physicists to browse hundreds of events in a single session without needing to refresh the page.

### DIFF
--- a/packages/phoenix-event-display/src/managers/three-manager/scene-manager.ts
+++ b/packages/phoenix-event-display/src/managers/three-manager/scene-manager.ts
@@ -458,13 +458,79 @@ export class SceneManager {
 
   /**
    * Clears event data of the scene.
+   * Properly disposes of all GPU resources (geometries, materials, textures)
+   * to prevent memory leaks during event switching.
    */
   public clearEventData() {
     const eventData = this.getEventData();
     if (eventData != null) {
+      this.disposeObject3D(eventData);
       this.scene.remove(eventData);
     }
     this.getEventData();
+  }
+
+  /**
+   * Recursively disposes of all GPU resources in an Object3D hierarchy.
+   * @param object The Object3D to dispose.
+   */
+  private disposeObject3D(object: Object3D) {
+    object.traverse((child: Object3D) => {
+      const mesh = child as Mesh | LineSegments | Line;
+
+      // Dispose geometry
+      if (mesh.geometry) {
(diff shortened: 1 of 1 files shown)


goyalpalak18:
Hi @EdwardMoyse @9inpachi , I’ve noticed the display can become unstable or crash after switching through multiple events—this is due to GPU memory not being released by Three.js automatically. I’ve implemented a recursive disposal helper in SceneManager to explicitly free up geometries, materials, and textures.

I've verified the fix with the Performance Monitor; VRAM now stays stable instead of climbing linearly. Ready for a review when you have a moment!

PR #792, opened 2026-01-26
Fix GPU Memory Leak in GLTF Geometry Import

remo-lab:
## The Bug
**File**: `packages/phoenix-event-display/src/managers/three-manager/import-manager.ts`  
**Function**: `ImportManager.loadGLTFGeometryInternal` (Lines 409-445)

When loading GLTF geometries, cloned geometries are collected in `materials[key].geoms` arrays, merged into a final geometry, but **intermediate geometries are never disposed**. This leaves orphaned WebGL buffers in GPU memory.

**Identical pattern** to the leak fixed in `phoenix-objects.ts` (commit ad79b442) but missed in import-manager.ts.

## The Fix
**Summary**: After `BufferGeometryUtils.mergeGeometries()`, added disposal loop for all intermediate geometries in the `materials` object.

**Files changed**: 1 (`import-manager.ts`)  
**Validation**: All 166 tests pass + linting clean

## Why This Matters
- **GPU memory leaks** on every GLTF load → browser crashes after multiple detector geometry loads
- **Physics workflows broken** (ATLAS/CMS detector visualization, control rooms, conferences)  
- **Silent failure**—no errors, just gradual memory exhaustion until OOM
- **Real production impact**—common workflow with complex detector geometries

## How to Verify
See `import-manager.ts` diff in this PR. To test locally:
1. Load any detector GLTF geometry multiple times
2. Monitor GPU memory via browser DevTools > Performance > Memory or `chrome://gpu`
3. **Before**: Memory grows unbounded  
   **After**: Memory stays stable after each load

## Impact
Eliminates crashes during long-running detector visualization sessions. Matches our existing cleanup pattern from `phoenix-objects.ts`. Zero render behavior change—just proper Three.js resource management.

### DIFF
--- a/packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
+++ b/packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
@@ -436,12 +436,17 @@ export class ImportManager {
             // Improve renderorder for transparent materials
             scene.remove(...scene.children);
             for (const val of Object.values(materials)) {
-              const mesh = new Mesh(
-                BufferGeometryUtils.mergeGeometries((val as any).geoms),
-                (val as any).material,
-              );
+              const intermediateGeoms = (val as any).geoms;
+              const mergedGeometry =
+                BufferGeometryUtils.mergeGeometries(intermediateGeoms);
+              const mesh = new Mesh(mergedGeometry, (val as any).material);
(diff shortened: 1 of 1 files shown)


EdwardMoyse:
Sorry, this was already fixed with #791

PR #822, opened 2026-03-12
feat: add time-aware EDM support and realistic animations (fixes #406)

### LINKED ISSUE #406: Time information and realistic animations
Dear developers,
As I saw the examples, event data of them don't have actual time information. So two questions: 

1. Is it correct that currently phoenix doesn't works anyhow with the right times and the animation is just a visual effect. 
2. Are there plans to work with timing information and provide realistic animations? 

The background of the question, at HallD in Jefferson Lab was this event viewer: 

https://halldweb.jlab.org/talks/2018/HDvis-Video/js/event.html#

It is not fancy but has real time information from the experiment. And our experience of using it was that it gave very valuable debugging information for people working on the real data reconstruction. At least it almost instantly revealed several bugs and lead to several enhancements. So it was proven to be valuable. 

Now I'm one of developers working on EIC Athena collaboration software. And we desperately want to start working on something to better represent our simulation/reconstruction data and for sure phoenix effort is something that we seriously consider to join.

deveshbervar:
- Add EventTime and EventMetadata interfaces to event-data-loader.ts
- Add optional getEventTime?() to EventDataLoader interface
- Extract event-level time (ns) in PhoenixLoader.buildEventData()
- Forward event time to AnimationsManager
- Add setEventTime(), getTimeProgress(), update() to AnimationsManager
- Add Event Time Progress slider to dat-gui UI
- Add tests for time extraction in phoenix-loader.test.ts

### DIFF
--- a/packages/phoenix-event-display/src/loaders/event-data-loader.ts
+++ b/packages/phoenix-event-display/src/loaders/event-data-loader.ts
@@ -7,20 +7,60 @@ import type {
 } from '../lib/types/event-data';
 
 /**
- * Event data loader for implementing different event data loaders.
+ * This file defines the interfaces used by event data loaders.
+ *
+ * Event data loaders are responsible for transforming raw event
+ * data into graphical objects and UI elements that can be rendered
+ * in the Phoenix event display.
+ */
+
+/**
+ * Metadata associated with an event.
+ * This information is typically shown in the UI to provide
+ * context about the event such as run number, event number,
+ * or time of recording.
  */
 export interface EventMetadata {
+  /** Label describing the metadata field (e.g. Run, Event). */
   label: string;
+
+  /** Value associated with the metadata label. */
   value: string | number;
+
+  /** Optional unit of the value (e.g. ns, GeV). */
   unit?: string;
-  time?: number; // ns
+
+  /** Optional time information in nanoseconds. */
+  time?: number;
 }
 
+/**
+ * Represents event-level timing information.
+ * The time value is expressed in nanoseconds and can be used
+ * by animation systems to synchronize event playback.
+ */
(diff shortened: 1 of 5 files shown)


EdwardMoyse:
Sorry, another long delay. So some quick comments, eventTime seems to be treated more as a `maxEventTime`?

The way this would work is if we had timing information added to e.g. hits. Then as the animation progressed, objects would grow and become visible. As it is right now, we assume that there is an expanding wave of particles coming from the origin in a sphere and hits appear as the sphere hits them. This assumption is completely wrong if the detector is not symmetric around an origin and if the particles do not come from the collision. An example from ATLAS would be for cosmic rays, where they come from top to bottom, and if we included hit level (or track level) timing information this could be visualised better.

So as it stands, I'm not really sure that this PR is going in the right direction. And in particular, if `eventTime` is `maxTime` then this is a function of the detector rather than the event.

(We also don't really use dat.gui any more)

deveshbervar:
Hi @EdwardMoyse,

Thank you for the detailed explanation — this is very helpful!

I understand now:
- eventTime as a global value is wrong — it's effectively maxTime 
  and belongs to the detector, not the event
- The right approach is per-object timing (hit.time, track.time) 
  so each object appears when the animation reaches its timestamp
- The current sphere animation assumption breaks for 
  non-symmetric detectors and cases like cosmic rays
- dat.gui is deprecated so I should remove that part

Would you prefer I:
1. Close this PR and open a new one focused only on adding 
   optional time fields to individual hits/tracks in the EDM?
2. Or continue in this PR with that direction?

I'm happy to start with just the EDM change (adding optional 
time field to hit/track objects) as a foundation, without 
touching the animation logic yet.

PR #841, opened 2026-03-20
fix(event-display): replace rainbow selection shader with glow outline

### LINKED ISSUE #839: Possibly revert the rainbow selection effect
In #721 we added a rainbow selection effect, which also showed the wireframe of the models. I think the outline glow effect we had before was a lot more visually pleasing and so would prefer reverting to this. What do you think @sponce?

deveshbervar:
Fixes #839

## What this does
Replaces the animated rainbow selection shader with a clean golden glow outline, addressing the visual concern raised in #839.

## Changes
- Replaced `SELECTION_FRAGMENT_SHADER` rainbow animation with a solid golden pulse (`vec3(1.0, 0.85, 0.3)`)
- Replaced `HOVER_FRAGMENT_SHADER` static blue with a pulsing blue (`vec3(0.3, 0.7, 1.0)`)
- Removed unused `RAINBOW_FUNCTION` constant
- Changed edge detection threshold to `1` for both hover and selection — removes the wireframe appearance

## What is preserved from #721
- Multi-object selection architecture
- EdgesGeometry rendering (more efficient than the old OutlinePass)
- Controls-based architecture
- Double-click orbit targeting
- Drag detection
- Programmatic selection API

## Note
Local build has a pre-existing `BrowserslistError: Unknown version 146 of chrome` issue on this machine unrelated to these changes.

### DIFF
--- a/packages/phoenix-event-display/src/managers/three-manager/effects-manager.ts
+++ b/packages/phoenix-event-display/src/managers/three-manager/effects-manager.ts
@@ -65,70 +65,26 @@ export class EffectsManager {
     }
   `;
 
-  /** Shared rainbow color function for reuse across shaders. */
-  private static readonly RAINBOW_FUNCTION = `
-    // Enhanced rainbow function - reused from main outline material
-    vec3 rainbow(float t) {
-      t = fract(t);
-      vec3 c = vec3(0.0);
-      
-      if (t < 0.16667) {
-        c = mix(vec3(1.0, 0.0, 1.0), vec3(1.0, 0.0, 0.0), t * 6.0);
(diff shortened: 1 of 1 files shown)


EdwardMoyse:
Argh. I'm sorry @deveshbervar - I had not realised that both you and @rx18-eng were working on similar topics. I'm really sorry about this - we should try not to overlap like this.

My comments to #837 apply here too - I definitely prefer the outline view. But interestingly, it seems like this is also not working on firefox / windows for me. So I am now wondering if this is caused by EdgesGeometry somehow, OR is specific to my computer at home.

image

PR #848, opened 2026-03-25
feat: add track kinematics panel with config-driven columns

### LINKED ISSUE #844: Track Kinematics Panel
##  Problem

  students load an event in Phoenix and see tracks in 3D but have no way to inspect their physics properties. in HYPATIA they get a
   table with pT, charge, eta, phi and use it to pick particles for invariant mass calculations. without this students can't do the 
  Z-path exercise in Phoenix. This is a missing HYPATIA feature.

  ## Plan

  I want to add a config driven kinematics panel. the idea is to define a KinematicsConfig with a column array where each column   
  has a getter function that extracts values from raw track data. this keeps it experiment agnostic (ATLAS default config but any  
  experiment can plug in its own).

  for the UI I plan to follow the existing CDK Overlay pattern (same as collections-info and geometry-browser) with a menu-toggle  
  button that creates an overlay via ComponentPortal. the table should be sortable by any column, have a min pT filter, and show   
  charge as colored numbers (red positive, blue negative) so students can quickly identify particle charges for lepton pair        
  selection.

  for 3D integration clicking a row should highlight that track so students can connect the numbers to the actual particle
  trajectory. eye icon to pan camera to the track. keyboard nav (arrow keys + enter) for quick browsing. I also want bidirectional 
  selection (hover track in 3D and table scrolls to that row) but this needs Line hover support in the selection manager which     
  doesnt exist yet so I will handle that separately.

  TSV export so students can save their data, collection dropdown when multiple track collections exist, and filtered/total count.

rx18-eng:
Closes #844 
## Summary                                                                                                                                                         
- Adds a track kinematics panel (sortable table showing pT, η, φ, charge, and   |p| for all tracks in the event)                                                - Config-driven architecture via KinematicsConfig so experiments define their own columns, units, and data extraction without modifying panel code          
- Includes ATLAS default config (ATLAS_KINEMATICS) with MeV→GeV conversion,charge from dparams[4]
                                                                                                                                        
## Why this matters for students                                                 
                                                                                
In masterclass exercises, students need to identify particles by inspecting track properties (high-pT tracks are Z/W decay products, charge determines) particle/antiparticle, and η tells you where in the detector the track goes.Currently students have to click individual tracks in 3D to see properties one at a time. This panel gives them the full picture at a glance with sorting and filtering so they can quickly spot the interesting tracks, similar to what
tools like HYPATIA provide.

  ## Features

  - Column sorting (click header), default pT descending — highest-energy tracks   first
  - pT threshold filter to cut low-momentum background                          
  - Collection dropdown for multiple track collections                          
  - Charge-colored index numbers and q values (red +, blue −)                     
  - TSV export for students to use data in their analysis
  - Keyboard navigation (arrow keys, enter to look at track)                      
  - Physics tooltips on column headers explaining what each quantity means      
  - Auto-refreshes on event change            
                                    
 **Adding kinematics for another experiment**
                                                                                
 ```
 const LHCb_KINEMATICS: KinematicsConfig = {                                   
    title: 'Track Momenta',
    columns: [
      { id: 'p', label: '|p|', unit: 'GeV', getter: (t) => t.momentum / 1000 },
      { id: 'charge', label: 'q', getter: (t) => t.charge ?? 0 },                   
      ],
  };                                                                            
        
``` 

https://github.com/user-attachments/assets/5a481df1-827c-4753-9882-7fae9e6583c9


EdwardMoyse:
I think this is absolutely beautiful, and really really nice. But... it is really duplicating what we already have in the collections panel isn't it? 

image

I completely agree for the masterclasses it is probably a much better and cleaner interface, but then I worry that the phoenix menu is becoming very cluttered. I think perhaps it could work however if we made a masterclass version of phoenix, either by adding a new option to the splashpage or if I create a new concrete implemention (with CERN hosted webpage) as with https://github.com/ATLAS-experiment/PhoenixATLAS ... another option would be to improve the existing collection panel, to make it more friendly (and even perhaps make it possible to limit the information it shows in "masterclass mode"?)

I think in any case we need a plan for how to incorporate all your new developments into a flexible, compact UI

Issue #474, opened 2022-07-06
Enabling calo cells causes a big performance hit


EdwardMoyse:
On my M1Pro mac, I can enable all event data except CaloCells in the ATLAS example:
https://hepsoftwarefoundation.org/phoenix/#/atlas
and I get 60 FPS. But as soon as I enable CaloCells this drops to 9FPS. This seems quite surprising so I wonder if there is a performance bug here.

PR #849, opened 2026-03-26
fix: use InstancedMesh for CaloCells to prevent WebGL crash (#474)

rx18-eng:
## Title
fix: use InstancedMesh for CaloCells to prevent WebGL crash (#474)

### Summary

This fixes the long-standing CaloCells performance issue (#474) where loading ATLAS events with calorimeter data (187,650 cells) would crash the browser. FPS dropped to ~1.8 and WebGL context was lost entirely, making the page unrecoverable.

I spent a good amount of time digging into this one. The root cause was straightforward but the fix touched a lot of surfaces. Phoenix was creating a separate THREE.Mesh (each with its own BoxGeometry + MeshPhongMaterial) for every single cell. That's 187K individual draw calls per frame. No GPU can handle that.

The fix replaces all of that with a single THREE.InstancedMesh (one geometry, one material, one draw call). Per-instance transforms and colors are stored in typed arrays. On my test machine (laptop), loading the same atlas.json that used to crash now runs at a stable 30 FPS with zero impact from CaloCells being enabled.

### What changed

- phoenix-objects.ts - New getCaloCellsInstanced() method that takes the entire cell array and builds one InstancedMesh with per-instance position, orientation, scale, and color
- object-type-registry.ts - CaloCells config now uses concatonateObjs: true so the full array is passed at once instead of per-cell
- scene-manager.ts - Filter and scale logic adapted for InstancedMesh. Filtering uses a scale-to-zero approach (hidden instances get a zero-scale matrix). Scale and filter coordinate through stored state so they don't overwrite each other
- selection-manager.ts - Raycaster now preserves instanceId for InstancedMesh hits, hover shows per-cell physics data in the info panel. OutlinePass selection is skipped since you can't outline individual instances
- controls-manager.ts - getObjectPosition() handles InstancedMesh via computeBoundingSphere()
- collections-info-overlay.component.ts - Visibility check reads the instance matrix instead of object.visible, internal fields (_instanceId, _position) excluded from the table
- phoenix-objects.test.ts - Test covering InstancedMesh creation, instance count, instanceId assignment, and shared uuid

### Issues I ran into along the way

- Frustum culling - Without calling computeBoundingSphere() after setting all instance matrices, Three.js uses the unit-box bounding sphere (radius ~0.87 at origin) and culls cells when panning. Took a bit to figure out why cells were disappearing
- Scale and filter conflicting - Both write to the same instance matrix buffer from original snapshots. If you scaled then filtered, the filter would restore un-scaled originals and silently lose the scale. Fixed by storing scale state on userData and re-applying it during filter restores
- UUID sharing - All 187K cells share one InstancedMesh uuid. Collections-info panel needed adaptation to check per-instance matrix state instead of per-object visibility
- lookAt in a batch loop - Can't call mesh.lookAt() per-instance. Used a temp Object3D with reset rotation before each lookAt() to extract quaternions for Matrix4.compose()

### Performance

Before: 187,000 draw calls, 187,000 scene graph children, 1.8 FPS (context lost on atlas.json)  
After: 1 draw call, 1 scene graph child, stable 30 FPS on laptop

https://github.com/user-attachments/assets/2751e6e5-3420-4e70-9833-d5b734089115

### DIFF
--- a/packages/phoenix-event-display/src/loaders/object-type-registry.ts
+++ b/packages/phoenix-event-display/src/loaders/object-type-registry.ts
@@ -89,7 +89,8 @@ export function getDefaultObjectTypeConfigs(): ObjectTypeConfig[] {
     },
     {
       typeName: 'CaloCells',
-      getObject: PhoenixObjects.getCaloCell,
+      getObject: PhoenixObjects.getCaloCellsInstanced,
+      concatonateObjs: true,
       cuts: [
         new Cut('phi', -pi, pi, 0.01),
         new Cut('eta', -5.0, 5.0, 0.1),
@@ -98,7 +99,7 @@ export function getDefaultObjectTypeConfigs(): ObjectTypeConfig[] {
       scaleConfig: {
         key: 'caloCellsScale',
         label: 'CaloCells Scale',
-        scaleMethod: 'scaleChildObjects',
+        scaleMethod: 'scaleInstancedObjects',
         scaleAxis: 'z',
       },
     },
--- a/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
+++ b/packages/phoenix-event-display/src/loaders/objects/phoenix-objects.ts
@@ -24,6 +24,8 @@ import {
   CanvasTexture,
   ShaderMaterial,
   DoubleSide,
+  InstancedMesh,
+  Matrix4,
 } from 'three';
 import { ConvexGeometry } from 'three/examples/jsm/geometries/ConvexGeometry.js';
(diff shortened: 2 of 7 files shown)


EdwardMoyse:
> Thanks for your review ! I renamed it to just CaloCell. For the hover - OutlinePass can't target individual instances so I used a color-based approach instead: hovered cell turns white, original color restores when you move away. Also fixed a bug where the info panel wasn't updating between cells (same object reference meant the hover-change check always returned false, now compares instanceId too).
> 
>  Recording.2026-04-02.050111.mp4

That looks very good - thanks!

PR #862, opened 2026-03-28
feat: animate pileup interactions sequenced by timestamp (#235)

### LINKED ISSUE #235: Animate pileup interactions?
Hello!

First of all, Phoenix is fantastic - thanks a lot for developing it! 

I have an idea for a feature request, and if we can develop a plan for how it can be implemented, I could possibly try to find some person-power to implement it.

At the HL-LHC we'll have precision-timing detectors in ATLAS and CMS that will help to separate the particles produced in up to 200 pp collisions that happen in a single bunch-crossing. It would be really fantastic to be able to animate this, just like the animate function now does. It's really not that much missing:
* Functionality to animate several pp interactions instead of just one
* If a comprehensive truth-record is fed in (we would have ~1000 particles), the interactions would be created at the time stored for each primary vertex, and the particles from each interaction would come out of that point.

Does this sound like an interesting feature? And would it be feasible to implement? If there are some pointers to the relevant parts of the code and a brief set of steps the experts foresee are needed, I could try to find a willing student, perhaps :)

Best,
Christian

deveshbervar:
Closes #235

## What this adds
Animates multiple pp interactions (pileup) from a single bunch crossing,
sequenced in time order using the timestamp stored on each interaction.
This is the foundational feature needed for HL-LHC precision-timing
detector visualisation in ATLAS and CMS.

## Demo

https://github.com/user-attachments/assets/94df574e-30ea-485f-92e6-82933a684750

## How it works
- Added `PileupInteraction` type with timestamp (ps), vertex (x,y,z), tracks and jets
- Added `PileupEvent` type containing an array of interactions
- `animatePileup()` sorts interactions by timestamp and maps each to a
  proportional delay, firing a vertex flash at the exact (x,y,z) position
- `_animateInteractionTracks()` handles per-interaction track animation
  when real tagged track data is provided
- New "Animate pileup" button added to the toolbar

## Testing
- All 179 existing tests passing, 0 failures
- Tested on ATLAS sample: 10 sequential vertex flashes visible along
  beam axis, each clearly separated by ~1 second

## Next steps for real HL-LHC data
Replace synthetic test data in `animate-event.component.ts` with real
JSON matching the `PileupEvent` format:
```json
{
  "interactions": [
    { "timestamp": 0,  "vertex": {"x":0,"y":0,"z":-45}, "tracks":[] },
    { "timestamp": 32, "vertex": {"x":0,"y":0,"z": 12}, "tracks":[] }
  ]
}
```

### DIFF
--- a/packages/phoenix-event-display/src/event-display.ts
+++ b/packages/phoenix-event-display/src/event-display.ts
@@ -15,6 +15,7 @@ import { URLOptionsManager } from './managers/url-options-manager';
 import type {
   PhoenixEventData,
   PhoenixEventsData,
+  PileupEvent,
 } from './lib/types/event-data';
 
 declare global {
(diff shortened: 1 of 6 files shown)


EdwardMoyse:
Sorry for taking so long to get back to you on this - I really wasn't quite sure what to say. Unfortunately I really think we should just close this PR - from the animation I can see that this is just fundamentally incorrect ... the pileup should be overlapping not played one after another. This is just not what is happening in reality. But I don't have time right now to try to work out if there is a correct way to do this - I suspect the timing differences from the different pileup will be imperceptible anyway.

Issue #861, opened 2026-03-27
feat: add event dataset browser with physics-aware filtering


rx18-eng:
### Problem

When loading multi-event files (PHYSLITE, CMS .ig, EDM4hep), Phoenix shows a simple dropdown with event keys ("Event 1", "Event 2"...). There's no way to know what's inside each event without loading it. Users click through events blindly looking for interesting topologies.

For PHYSLITE files with 100+ events, or masterclass datasets with 50 events, this wastes significant time. Students spend more time clicking "next" than analyzing physics.

### Proposed Solution

An event dataset browser panel that pre-scans all loaded events and shows a sortable, filterable physics summary.

Core features:
- Pre-scan events on load, build summary index (object counts per collection, MET, total energy)
- Sortable table (click column header to sort by # jets, # muons, MET, etc.)
- Filter controls (e.g. "events with >= 2 muons", "MET > 20 GeV")
- Click a row to load that event
- Keyboard navigation (arrow keys to browse, Enter to load)
- Active event highlighted in the table

Design principles:
- Config-driven columns and filters (experiment-agnostic, works for ATLAS/CMS/LHCb/FCC)
- No new dependencies (reuses existing Angular patterns)
- Works with any data format that Phoenix supports

Example column config:

```ts
{
  columns: [
    { key: 'eventNumber', label: 'Event' },
    { key: 'Jets', label: 'Jets', countCollection: true },
    { key: 'Tracks', label: 'Tracks', countCollection: true },
    { key: 'Muons', label: 'Muons', countCollection: true },
  ],
  filters: [
    { field: 'Muons', operator: '>=', label: 'Min Muons' },
    { field: 'MET', operator: '>=', label: 'Min MET (GeV)' },
  ]
}
```

### Use Cases

- Students in masterclasses filtering to "2 muon events" to find Z candidates instead of clicking blindly through 50 events
- Researchers finding events with specific topologies (high MET, many jets, isolated leptons)
- Developers verifying that their loader produces the expected object counts across events
- Presenters picking the cleanest event for a talk or demo

### Additional Context

Currently the event selector is a `<select>` dropdown (event-selector.component.ts) with no search, filter, or preview capability. Events are already all held in memory (eventsData in EventDisplay), so pre-scanning requires no additional data loading.

No keyboard shortcuts exist for event navigation (next/prev). This feature would also add arrow key and Enter support for event browsing.

rx18-eng:
[...]

PR #868, opened 2026-03-30
feat: add event dataset browser with physics-aware filtering (#861)

rx18-eng:
### Summary

- added a new Event Browser panel that pre-scans all loaded events and shows a summary table with object counts per collection type, MET, and event/run numbers
- the table is sortable and filterable (column filters with >=, <=, =, MET minimum, and search by event number)
- keyboard nav works with arrow keys in the table + Shift+Left/Right for prev/next event globally

columns are ordered with reconstructed physics objects first (Jets, Muons, Tracks, etc) and detector-level stuff (CaloCells, Hits) pushed to the right. zero values are dimmed so you can quickly scan which events have what. there's a separate "Reco" column that only counts reconstructed objects for easier comparison between events.

follows the same CDK overlay pattern as Collections Info, config-driven via showEventBrowser in UIMenuConfig.

search by event number is there per Edward's feedback. the UI is independent of the loading strategy so lazy-loading can be added later without changing the browser itself (per Sebastien's note on memory).

### Test plan

- load a multi-event file (atlas.json from assets/files/event_data/ works)
- open Event Browser from toolbar
- check that reconstructed columns show before detector-level ones
- try sorting, adding filters, searching by event number
- click load button and check the spinner + loaded event highlight
- Shift+Right/Left to navigate events globally

https://github.com/user-attachments/assets/0396390c-e422-4605-9385-fb50fabb65e4

### DIFF
--- a/packages/phoenix-event-display/src/event-display.ts
+++ b/packages/phoenix-event-display/src/event-display.ts
@@ -16,6 +16,10 @@ import type {
   PhoenixEventData,
   PhoenixEventsData,
 } from './lib/types/event-data';
+import {
+  buildEventSummaries,
+  type EventSummary,
+} from './helpers/event-summary';
 
 declare global {
   /**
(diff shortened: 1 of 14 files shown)


sponce:
No it's all fine to me. Just wanted the feelings of @EdwardMoyse before merging.

rx18-eng:
@EdwardMoyse would love to know you thoughts on this !

rx18-eng:
@EdwardMoyse shall we proceed merging this if you like the feature !

EdwardMoyse:
Sorry for the delay! This does look really great!

PR #876, opened 2026-04-08
feat: add eta-phi energy map panel

rx18-eng:
## Body

Adds an interactive eta-phi energy heatmap panel to the UI menu.

Bins calorimeter energy (CaloCells + CaloClusters) into a 100x63 grid over eta [-5, 5] and phi [-pi, pi], and renders it on a Canvas 2D surface with a sequential warm colormap (blue to red to white).

### What it does:

- Heatmap of calorimeter energy deposits in eta-phi space
- Overlay markers for Jets (cone ellipses), Muons, Electrons, Photons, Tracks (dots), and MET (arrow)
- Hover tooltips showing energy in GeV, eta, phi for each bin, or the object label for markers
- Clicking a marker highlights and zooms to that object in the 3D view
- Log/linear scale toggle for the color mapping
- Overlays on/off toggle
- Color bar showing the energy scale
- Auto-rebuilds when switching events

The config (EtaPhiConfig) is generic and not hardcoded to ATLAS. Other experiments can pass different eta/phi ranges and bin counts. Tracks derive eta/phi from dparams when direct fields aren't available.

Used native canvas event listeners instead of Angular template bindings because CDK overlay portals don't play well with Angular's change detection for rapid mouse events. Tooltip is drawn directly on the canvas for the same reason.

Follows the same CDK Overlay + ComponentPortal pattern as the other panels (collections-info, geometry-browser, etc).

https://github.com/user-attachments/assets/1f4c3415-c969-49bc-806c-ca8158259567

### DIFF
--- a/packages/phoenix-event-display/src/helpers/eta-phi-config.ts
+++ b/packages/phoenix-event-display/src/helpers/eta-phi-config.ts
@@ -1,0 +1,25 @@
+/** Configuration for the eta-phi energy map panel. */
+export interface EtaPhiConfig {
+  /** Eta axis range. Defaults to [-5, 5]. */
+  etaRange?: [number, number];
+  /** Phi axis range in radians. Defaults to [-pi, pi]. */
+  phiRange?: [number, number];
+  /** Number of bins along eta. Defaults to 100. */
+  etaBins?: number;
+  /** Number of bins along phi. Defaults to 63. */
+  phiBins?: number;
+  /** Minimum energy (GeV) to display. Defaults to 0 (hide noise). */
+  energyThreshold?: number;
+  /** Use logarithmic color scale. Defaults to true. */
+  logScale?: boolean;
+}
+
+/** Default ATLAS-style eta-phi configuration. */
+export const DEFAULT_ETA_PHI_CONFIG: Required<EtaPhiConfig> = {
+  etaRange: [-5, 5],
+  phiRange: [-Math.PI, Math.PI],
+  etaBins: 100,
+  phiBins: 63,
+  energyThreshold: 0,
+  logScale: true,
+};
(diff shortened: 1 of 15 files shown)


EdwardMoyse:
Wow. This is very cool ... please bear with me though, I've had a quick play with it and I love the concept, but it will take me a little while to review it. I'm a bit puzzled why clusters appear and disappear if I turn log scale off/on for example, but I need to really read the code properly to be able to comment.

rx18-eng:
thanks for checking it out!
                             
the clusters aren't actually disappearing, its a dynamic range thing. calorimeter energies can span 3-4 orders of magnitude, so on linear scale the low energy cells end up with colors that are basically the same as the dark background. log scale compresses that range which is why everything becomes visible again when you toggle it back.

that's why log is the default (its the standard for lego plots). but I can see how it'd be confusing for students if things seem to vanish when they flip that toggle. I could add a minimum brightness floor so even the coldest cells stay faintly visible on linear, if you think that'd help?

take your time with the full review, happy to address everything together.

rx18-eng:
@sponce would also like to know your thoughts on this !

EdwardMoyse:
Unfortunately this has now picked up conflicts - can you take a look?

rx18-eng:
Sure I will be solving them by today

rx18-eng:
@EdwardMoyse shall we proceed merging this ?

sponce:
Finally found a bit of time to look at this. Looks great for Atlas, but it is specific to detectors having certain data (hardcoded names at https://github.com/HSF/phoenix/pull/876/changes#diff-54e5196b2c8a0cd99ce33ff16e5ef4dfc5e754e594b7b13dc48e4f24e0495249R213).
In LHCb case, this does not really makes sense and anyway the plot is empty. So we should probably remove the button from the menu bar for LHCb

rx18-eng:
Makes sense I will be updating this today !

EdwardMoyse:
Happy to accept this once the agreed change is made!

rx18-eng:
@EdwardMoyse Let me know if any other changes are required :)

rx18-eng:
@EdwardMoyse shall we proceed merging this !

PR #877, opened 2026-04-09
perf(phoenix-ng): isolate animation loop outside NgZone

GaneshPatil7517:
Run the Three.js animation loop outside Angular's zone to prevent unnecessary change detection cycles (~60/sec) caused by requestAnimationFrame.

Override EventDisplayService.init() to wrap super.init() inside NgZone.runOutsideAngular(), ensuring the setAnimationLoop callback and all its descendant microtasks execute outside Angular's zone.

Angular UI still re-enters the zone for user-driven events (click, input) since Zone.js patches those globally.

Ref: HSF/phoenix#827


EdwardMoyse:
I tested this locally - it is really great! Thank you very much.

GaneshPatil7517:
> I tested this locally - it is really great! Thank you very much.

thank you @EdwardMoyse

Issue #886, opened 2026-04-18
refactor: EffectsManager missing state model, hover color not configurable, no experiment override API


deveshbervar:
## Context

PR #837 improved `EffectsManager` significantly — OutlinePass for
selection silhouettes, Fresnel shader for jets, `setSelectionColor()`
for experiment customization. Strong foundation.

Four gaps remain that block further extensibility, all verified
against current `main` after #837 merged.

---

## Gap 1 — GLSL embedded as TypeScript strings, no CI validation

`VERTEX_SHADER` and `HOVER_FRAGMENT_SHADER`
are `private static readonly` TypeScript string constants containing
raw GLSL. A syntax error only surfaces at WebGL runtime — never in
`yarn test:ci`. Search of `.github/workflows/` for `glslang`,
`glsl-loader`, `glsl-validate` returns zero results.

The Fresnel shader added in PR #837 (`phoenix-objects.ts`) follows
the same pattern — no shader in the codebase has a validation path.

---

## Gap 2 — No formal state model

The render path branches on implicit checks:

- Line 148: `if (this.selectedObjectsSet.size > 0)`
- Line 157: `else if (this.hoverOutline)`
- Line 219: `if (this.selectionOutlinePass && this.selectedObjectsSet.size > 0)`
- Line 268: `this.selectionOutlinePass.enabled = this.selectedObjectsSet.size > 0`

Search for `EffectsState` returns zero results — no enum, no typed
state machine exists. Adding a new visual state — `DIMMED` for pileup
animation (PR #862) or `HIGHLIGHTED` for masterclass mode — requires
finding and updating every implicit branch with no compiler guidance.

---

## Gap 3 — Hover color hardcoded, selection color configurable

`setSelectionColor(color: number)` exists at line 328 — experiments
can override the amber selection outline. Hover color is hardcoded
at line 71:

```glsl
vec3 color = vec3(0.2, 0.6, 1.0);
```

No `setHoverColor()` exists. An experiment with blue-colored objects
(e.g. LHCb calorimeter deposits) cannot change hover color without
modifying core code — inconsistent with the configurable selection
color pattern established by #837.

---

## Gap 4 — No experiment effect override API

Search for `registerEffect`, `effectOverride`, `override` in
`effects-manager.ts` returns zero results. Search of all experiment
pages in `packages/phoenix-ng/projects/phoenix-app/src/app/sections/`
for `effectsManager.` calls returns zero results.

[...]

PR #889, opened 2026-04-23
feat(effects-manager): add EffectsState enum and setHoverColor() (#886)

deveshbervar:
## Summary

Phase 1 (part 1) of the refactoring proposed in #886.

Introduces the `EffectsState` enum with immediate usage in hover and
selection state tracking, and adds `setHoverColor()` to make hover
color configurable — matching the existing `setSelectionColor()` API.
Both changes are introduced together so the enum has immediate,
meaningful usage.

## What this adds

**EffectsState enum** — typed visual state management:

```typescript
export enum EffectsState {
  DEFAULT = 'DEFAULT',
  HOVERED = 'HOVERED',
  SELECTED = 'SELECTED',
  HIGHLIGHTED = 'HIGHLIGHTED',
  DIMMED = 'DIMMED',
}
```

Used via `private currentState: EffectsState`:
- `selectObject()` → `SELECTED`
- `deselectObject()` → `DEFAULT` when no selections remain
- `clearAllSelections()` → `DEFAULT`
- `setHoverOutline()` → `HOVERED` or `DEFAULT`

**setHoverColor(color: number)** — mirrors `setSelectionColor()`:

```typescript
effectsManager.setHoverColor(0xff6600); // orange for LHCb
```

Updates the active hover outline immediately if one exists.
Hover color stored as RGB in `_hoverColor` and passed to
`HOVER_FRAGMENT_SHADER` via uniforms — replacing the previously
hardcoded `vec3(0.2, 0.6, 1.0)`.

## Why

From #886 Gap 2: no typed state model existed — the render path relied
on implicit boolean checks with no compiler guidance.

From #886 Gap 3: `setSelectionColor()` existed but hover color was
hardcoded. Experiments with blue objects (e.g. LHCb) had no way to
customize hover color without modifying core code.

## Scope

- `effects-manager.ts` only
- No render path changes
- No shader file extraction
- All existing tests pass

Part of #886 — Phase 1, part 1.

### DIFF
--- a/packages/phoenix-event-display/src/managers/three-manager/effects-manager.ts
+++ b/packages/phoenix-event-display/src/managers/three-manager/effects-manager.ts
@@ -15,6 +15,24 @@ import { OutlinePass } from 'three/examples/jsm/postprocessing/OutlinePass.js';
 import { Pass } from 'three/examples/jsm/postprocessing/Pass.js';
 
 /**
+ * Represents the possible visual states of objects managed
+ * by EffectsManager. Provides typed state management instead
+ * of implicit boolean checks scattered across the render path.
+ */
+export enum EffectsState {
+  /** Default state of the object. */
+  DEFAULT = 'DEFAULT',
+  /** State when the object is being hovered over. */
+  HOVERED = 'HOVERED',
+  /** State when the object is selected. */
+  SELECTED = 'SELECTED',
+  /** State when the object is highlighted. */
+  HIGHLIGHTED = 'HIGHLIGHTED',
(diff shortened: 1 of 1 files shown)


sponce:
Looks good to me. There is still a number of unrelated comment changes, mostly very welcome rewording but still unrelated, so which should probably go in a different commit at least (we can keep them in the MR)

deveshbervar:
Thanks @sponce I'll separate the comment rewording 
into its own commit to keep the functional changes clean.

EdwardMoyse:
Thanks - this looks good.

PR #895, opened 2026-05-15
feat(phoenix-ng): add NotificationService with 4 severity levels and configurable auto-dismiss

deveshbervar:
## Summary

Adds `NotificationService` as part of Work Area 1 (WA1).

WA1 arc:
-  Step 1 — IO error handling with try/catch (#879, merged)
-  Step 2 — NotificationService (this PR)
-  Step 3 — UI component that subscribes and shows toasts(next part)

## What this adds

**`notification.service.ts`** — Injectable Angular service:

```typescript
notificationService.success('File loaded successfully.');
notificationService.info('Loading event data...');
notificationService.warning('File may be malformed.');
notificationService.error('Could not parse event file.');
```

Follows the same `ActiveVariable` pattern as `ErrorMessageService`.
Components subscribe via `subscribeToNotifications()`.

Default auto-dismiss durations:
- `success` — 5000ms
- `info` — 5000ms  
- `warning` — 8000ms
- `error` — 0ms (manual dismiss required)

Custom duration supported via optional second parameter.

**`notification.service.component.test.ts`** — 6 tests:
- All 4 severity levels emit correct message and severity
- Default durations per severity level
- Custom duration override
- Multiple subscribers supported

## Why

PR #879 added try/catch to route IO errors through InfoLogger.
This PR adds the service layer that allows any Phoenix component
to emit user-visible notifications programmatically.

The next step (separate PR) will add the UI component that
subscribes to this service and renders toasts via MatSnackBar.

## Scope

- 2 new files in `phoenix-ui-components/lib/services/`
- 1 line added to `index.ts` for export
- No changes to existing files beyond the export
- 6 tests passing locally

### DIFF
--- a/packages/phoenix-ng/projects/phoenix-ui-components/lib/services/index.ts
+++ b/packages/phoenix-ng/projects/phoenix-ui-components/lib/services/index.ts
@@ -1,3 +1,4 @@
 export * from './event-display.service';
 export * from './extras/event-data-import';
 export * from './extras/attribute.pipe';
+export * from './notification.service';
--- a/packages/phoenix-ng/projects/phoenix-ui-components/lib/services/notification.service.component.test.ts
+++ b/packages/phoenix-ng/projects/phoenix-ui-components/lib/services/notification.service.component.test.ts
@@ -1,0 +1,53 @@
+import { NotificationService } from './notification.service';
+
+describe('NotificationService', () => {
+  let service: NotificationService;
+  let received: any[];
+
+  beforeEach(() => {
+    service = new NotificationService();
+    received = [];
+    service.subscribeToNotifications((n) => received.push(n));
+  });
+
+  it('should emit success notification', () => {
+    service.success('Operation complete');
+    expect(received[0].severity).toBe('success');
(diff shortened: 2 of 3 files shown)


sponce:
Nice service, now I would not merge it before it's effectively used. By the way, I see you want to use MatSnackBar to display the messages, that would be a new dependency to my knowledge. Don't we already have a way to display messages ? @EdwardMoyse ?

deveshbervar:
Thanks @sponce 

Understood on merging before use — I'll integrate it with 
its first usage before it merges.
On the existing system: InfoLogger writes to the persistent 
Info Panel (requires the panel to be open). The 
NotificationService is designed as a complementary transient 
layer — auto-dismissing toasts that appear immediately.

On MatSnackBar:- @angular/material is already a dependency 
in phoenix-ui-components — so no new dependency is introduced.

@EdwardMoyse - happy to hear your thoughts too.

EdwardMoyse:
Hi all, apologies for not responding earlier - I was travelling last week. I think this is a very nice idea!

PR #917, opened 2026-05-31
fix: Prevent autoloader loop crash on individual file fetch errors

lavjeetrai:
### What changed?
Wrapped the `fetchAndEmit(fileUrl)` call inside a `try/catch` block within the directory polling loop. 

### Why is this needed?
Previously, if a single file failed, the uncaught error would break the entire `for` loop. This caused the autoloader to completely ignore any remaining new files in that batch. 

By isolating the fetch step, a failed file is now safely caught and passed to the `onError` handler, allowing the loop to continue processing the rest of the files without interruption.

### DIFF
--- a/phoenix/packages/phoenix-event-display/src/helpers/event-autoloader.ts
+++ b/phoenix/packages/phoenix-event-display/src/helpers/event-autoloader.ts
@@ -113,10 +113,17 @@ export class EventAutoloader {
 
     const newFiles = this.parseApacheListing(html, url);
 
-    for (const fileUrl of newFiles) {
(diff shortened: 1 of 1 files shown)


lavjeetrai:
@EdwardMoyse can u review this pr when u get time ?

EdwardMoyse (approved):
Looks good. Thanks!

lavjeetrai:
@EdwardMoyse ,Thanks so much for the review and for merging this! I'd love to continue contributing. Are there any specific issues or upcoming features you'd recommend I look at next? Happy to take on something a bit more challenging if you have anything in mind.

PR #835, opened 2026-03-18
feat: add configurable masterclass panel with event bus for external integration

rx18-eng:
Closes #824 #826

  Adds a configurable masterclass panel for tagging particles and computing     
  invariant masses, designed to be reusable across experiments.

  ## What it does

  Students can select tracks from any collection, tag them as particle types,   
  see kinematic values in a table, and compute invariant masses. Results can be 
  recorded across events and exported as a text file.

  ## Architecture

  The panel is **experiment-agnostic**. All experiment-specific details
  (particle types, masses, hints, event classification) are defined in a        
  `MasterclassConfig` interface and passed as an input — not hardcoded.

  - **`MasterclassConfig`** — defines particle tags (id, symbol, color, mass),  
  educational hints, and a classification callback
  - **`ATLAS_MASTERCLASS_CONFIG`** — default config for ATLAS Z-path exercises  
  (e±/μ±/γ, Z/Higgs/J/ψ hints)
  - Adding a new experiment's masterclass (e.g. LHCb D0 → Kπ) requires only a   
  config object — no component changes

  ## Integration API

  A lightweight event bus (`on`/`emit`) is added to `EventDisplay`, enabling    
  external frameworks to subscribe to live data:

  ```
typescript
  eventDisplay.on('particle-tagged', (data) => { /* tag, uuid, fourMomentum, pT,
   eta, phi */ });
  eventDisplay.on('result-recorded', (data) => { /* eventType, mass,
  particleCount */ });
```

  This allows external tools (e.g. histogram frameworks) to react to user       
  actions without modifying Phoenix.

  ## Files

  Core library (phoenix-event-display):
  - helpers/invariant-mass.ts — generic 4-momentum math +
  MasterclassConfig/ParticleTagDef types + ATLAS default config
  - event-display.ts — on(event, callback) / emit(event, data) event bus        

  ## UI components (phoenix-ui-components):
  - masterclass-panel/ — toolbar toggle + overlay panel, config-driven via      
  @Input()
  - ui-menu.component — threads masterclassConfig input to panel

  App (phoenix-app):
  - atlas.component — passes ATLAS_MASTERCLASS_CONFIG to ui-menu

  ## Test plan

  - Load ATLAS section with default JiveXML event
  - Open masterclass panel (atom icon), select tracks, tag particles
  - Verify invariant mass computation and hints display
  - Record results and export file
  - Verify no console errors on LHCb/other sections

  

[...]

Issue #915, opened 2026-05-29
Add new masterclass example to the landing page


EdwardMoyse:
A lot of features have been added recently, related to masterclasses. For masterclasses it would be good to add a new example to the landing page with simplified UI (i.e. a reduced Menu bar at the bottom).

rx18-eng:
Hey @EdwardMoyse  great to  hear that we are finally including this in the ui :) . I will surely work on this .

EdwardMoyse:
Yeah sorry it took so long, but I finally found a bit of time to work on Phoenix again!

rx18-eng:
No worries  I will start working on this and my new ideas from today and will also work on all the open issue regarding release 4.1.0

Issue #914, opened 2026-05-29
Add documentation about PHYSLITE ROOT file loader


EdwardMoyse:
See https://github.com/HSF/phoenix/pull/836

Issue #913, opened 2026-05-29
Add documentation about Event Dataset browser


EdwardMoyse:
See https://github.com/HSF/phoenix/pull/868

PR #920, opened 2026-06-01
feat: add masterclass panel and ATLAS masterclass landing page (close…

rx18-eng:
## Body

closes #835 and #915 in one PR. #835 was stale (branch drifted too far to rebase practically), so i rebuilt the masterclass code on a fresh branch from main instead.

### what's in

- masterclass tagging panel. opt-in via `uiConfig.showMasterclassPanel` (default false so `/atlas` etc stay unchanged). tag particles, live invariant mass, results history.
- `/atlas-masterclass` route with a reduced bar. same atlas detector + event as `/atlas`. drops vr, ar, animate, clipping, overlay, share, make-picture. more info menu keeps info panel, event browser, eta-phi, kinematics (drops collections info + geometry browser).
- landing card between atlas and lhcb pointing to the new route.
- `on()` / `emit()` event bus on `EventDisplay` (addresses part of #826). standard events fired:
  - `particle-tagged`
  - `particle-untagged`
  - `result-recorded`

### not in (intentionally)

- histogram. #843 stays separate, will rebase + follow up.
- LHCb / CMS masterclass configs. sebastien mentioned looping in bcouturi for LHCb so leaving that for later.
- docs (separate, addressing #911 #913 #914)

https://github.com/user-attachments/assets/b4403b4f-137a-4c9e-b151-a164d3789442

### DIFF
--- a/packages/phoenix-event-display/src/event-display.ts
+++ b/packages/phoenix-event-display/src/event-display.ts
@@ -45,6 +45,8 @@ export class EventDisplay {
   private onEventsChange: ((events: any) => void)[] = [];
   /** Array containing callbacks to be called when the displayed event changes. */
   private onDisplayedEventChange: ((nowDisplayingEvent: any) => void)[] = [];
+  /** Generic event bus for integration with external frameworks. */
+  private eventBus: Map<string, Set<(data: any) => void>> = new Map();
   /** Three manager for three.js operations. */
   private graphicsLibrary: ThreeManager;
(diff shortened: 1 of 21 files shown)


rx18-eng:
@EdwardMoyse Let me know your thoughts on this !

EdwardMoyse:
This looks great. I think we should get this in now, then I can share it with some masterclass folk. We can implement any feedback for 4.1.0

PR #933, opened 2026-06-12
refactor(effects-manager): extract GLSL shaders to .glsl files via raw-loader (#886)

deveshbervar:
## Summary

Phase 2 (PR 2a of 2) of the refactoring proposed in #886.

Extracts the two inline GLSL shader strings from `effects-manager.ts` into dedicated `.glsl` files,
loaded at build time via `raw-loader`.

## What changed

**New files:**
- `shaders/hover-vertex.glsl` — vertex shader extracted from TS string
- `shaders/hover-fragment.glsl` — fragment shader extracted from TS string
- `src/glsl.d.ts` — TypeScript declaration for `.glsl` imports
- `src/tests/helpers/glsl-mock.js` — Jest mock for `.glsl` files

**Modified files:**
- `configs/webpack.conf.js` — added `raw-loader` rule for `.glsl`
- `configs/jest.conf.js` — added `moduleNameMapper` for `.glsl` in tests
- `effects-manager.ts` — replaced inline strings with ES6 imports
- `package.json` — added `raw-loader` as devDependency

## Why

From #886 Gap 1: GLSL embedded as TypeScript strings has no validation path — a syntax error only surfaces at WebGL
runtime, never in `yarn test:ci`.

This extraction is the prerequisite for PR 2b which adds `glslangValidator` CI validation.

## Scope

- Zero behavior change — shaders are identical
- All 25 effects-manager tests pass
- Part of #886 — Phase 2, PR 2a of 2

### DIFF
--- a/packages/phoenix-event-display/src/managers/three-manager/effects-manager.ts
+++ b/packages/phoenix-event-display/src/managers/three-manager/effects-manager.ts
@@ -13,6 +13,8 @@ import {
 import { RenderPass } from 'three/examples/jsm/postprocessing/RenderPass.js';
 import { OutlinePass } from 'three/examples/jsm/postprocessing/OutlinePass.js';
 import { Pass } from 'three/examples/jsm/postprocessing/Pass.js';
+import VERTEX_SHADER from './shaders/hover-vertex';
+import HOVER_FRAGMENT_SHADER from './shaders/hover-fragment';
 
 /**
  * Represents the possible visual states of objects managed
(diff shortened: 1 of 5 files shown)


packages/phoenix-event-display/src/managers/three-manager/effects-manager.ts
```diff
@@ -14,6 +14,34 @@ import { RenderPass } from 'three/examples/jsm/postprocessing/RenderPass.js';
...
+  uniform float opacity;
+  uniform float colorR;
+  uniform float colorG;
+  uniform float colorB;
+
+  void main() {
+    gl_Position = projectionMatrix * modelViewMatrix * vec4(position, 1.0);
+  }
+  `;
+
```
sponce: I'm not really sure I'm qualified to comment on this MR, as I have actually no clue how shaders work.
So be indulgent if my question is stupid, but I'm just a bit surprised that the "extraction of GLSL shaders' as the title says only results is having the shader code duplicated in 2 places, this one and the new nice glsl file.  Is there no way to have it only in the glsl file ? What if the 2 are not in sync ?

deveshbervar:
Hi @sponce,
My apologies for the back and forth on this PR. I have some questions regarding the best approach for shader extraction.
I will discuss it with @EdwardMoyse and come back with a cleaner/new PR. For now, I’m closing this one.
Thank you for your earlier review!

PR #931, opened 2026-06-07
Enhance UI components stacking and panel behavior

lavjeetrai:
## Overview
This PR improves the behavior and usability of floating data panels in the Phoenix UI and also addresses issue #925.

https://github.com/user-attachments/assets/32367184-d8a1-46a7-8f3e-6ebcbcaea2d2

## Key Changes

- Made `Collections Info`, `Event Browser`, and `Geometry Browser` fully independent floating panels.
- Added better default positioning for panels to avoid overlap and improve layout consistency.
- Added scrolling constraints (`max-width: 85vw`, `max-height: 40vh`) for large tables instead of letting panels overflow the screen.
- Fixed z-index issues so Angular Material menus and tooltips correctly appear above overlays.

PR #934, opened 2026-06-12
refactor to ImportManager to significantly improve performance,

lavjeetrai:
## Description

This PR introduces a major refactor to `ImportManager` to significantly improve performance, memory management, and overall code architecture for importing 3D assets. It addresses several bottlenecks and enforces stricter TypeScript type safety.

### Key Improvements

- **Lazy Instantiation & Caching:** Loaders (`GLTFLoader`, `OBJLoader`, `DRACOLoader`) are now lazily instantiated and cached to prevent upfront CPU/Memory spikes when the module is loaded.
- **GPU Memory Leak Prevention:** Added a `disposeMaterial` helper that reliably disposes of intermediate materials and their deeply nested textures.
- **Geometry Merging for Draw Call Optimization:** GLTF meshes sharing the same material are now merged using `BufferGeometryUtils.mergeGeometries`. This drastically reduces WebGL draw calls, with a safe fallback to individual meshes if attribute mismatches occur.
- **Unified Asynchronous Flow:** Removed legacy callback anti-patterns (e.g., inside `parsePhnxScene`) in favor of modern `async/await` and Promises, making the codebase much easier to trace and maintain.
- **Dynamic JSZip Imports:** The `jszip` library is now loaded dynamically (`await import('jszip')`) only when a `.zip` file is detected, reducing the initial JavaScript bundle size.
- **Resilient ZIP Extraction:** Upgraded from `Promise.all` to `Promise.allSettled` when extracting multiple files from a `.zip`, ensuring that a single corrupted file won't reject the entire valid payload.
- **Improved Type Safety:** Eliminated arbitrary `any` assertions across material property checks and geometry arrays, leveraging strict type guards (e.g., `MeshStandardMaterial | MeshBasicMaterial | MeshPhongMaterial`) to prevent runtime errors.

###  Testing Instructions
1. Load an `.obj` file and ensure it renders with the correct flat/double-sided properties.
2. Load a `.gltf` or `.glb` scene (with and without DRACO compression) and confirm textures/geometries are correctly preserved.
3. Import a `.zip` file containing multiple geometries and verify it unpacks and renders properly.
4. Load a `.phnx` event scene and verify that both the event data and geometries are initialized without callback errors.

### DIFF
--- a/packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
+++ b/packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
@@ -13,13 +13,17 @@ import {
   Vector3,
   Matrix4,
   REVISION,
+  ColorRepresentation,
+  Texture,
+  BufferGeometry,
+  MeshStandardMaterial,
+  MeshBasicMaterial,
 } from 'three';
 import { OBJLoader } from 'three/examples/jsm/loaders/OBJLoader.js';
 import { GLTFLoader } from 'three/examples/jsm/loaders/GLTFLoader.js';
 import { DRACOLoader } from 'three/examples/jsm/loaders/DRACOLoader.js';
 import type { GeometryUIParameters } from '../../lib/types/geometry-ui-parameters';
 import * as BufferGeometryUtils from 'three/examples/jsm/utils/BufferGeometryUtils.js';
-import JSZip from 'jszip';
 
 /**
  * Manager for managing event display's import related functionality.
@@ -31,6 +35,13 @@ export class ImportManager {
   private EVENT_DATA_ID: string;
   /** Object group ID containing detector geometries. */
   private GEOMETRIES_ID: string;
+
+  /** Cached GLTFLoader instance */
+  private gltfLoader?: GLTFLoader;
+  /** Cached OBJLoader instance */
+  private objLoader?: OBJLoader;
+  /** Cached ObjectLoader instance */
+  private objectLoader?: ObjectLoader;
 
   /**
    * Constructor for the import manager.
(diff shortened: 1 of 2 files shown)


lavjeetrai:
@sponce @EdwardMoyse hey guys, whenever you have a minute, could you take a look at this refactor? Cleaned up some of the GLTF loading logic.

sponce (approved):

sponce:
Just a tiny question : shall we squash the 3 commits, for the history to be cleaner ? Or feel free to rewrite the history yourself and push the new version

lavjeetrai:
I've squashed the commits into a single commit. Let me know if everything else looks good to go.

Also, PR https://github.com/HSF/phoenix/pull/931 has been waiting for more than a week. Since the @EdwardMoyse seems to be offline, could you advise on whether there are any remaining blockers to getting it merged?

packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
```diff
@@ -151,207 +213,140 @@ export class ImportManager {
         child.name = object3d.name;
         child.userData = object3d.userData;
         child.userData.size = this.getObjectSize(child);
-        // Use the new material
-        if (child.material instanceof Material) {
-          child.material.dispose();
-          child.material = material2;
+
+        if (child.material) {
```
sponce, line 217: May be a stupid question : why do we not check anymore the type ? disposeMaterial needs a Material

sponce (approved):

packages/phoenix-event-display/src/managers/three-manager/import-manager.ts
lavjeetrai: In Three.js, a mesh's `material` can be either a single `Material` or an array (`Material[]`). 

The previous `instanceof Material` check evaluated to `false` for arrays, skipping disposal and causing memory leaks. By removing the strict type check, we safely pass both types to the new `disposeMaterial` helper, which is designed to iterate through and dispose of them correctly.
Does this approach make sense to you?

sponce: ok, thanks for the details. All fine then

PR #945, opened 2026-06-27
Update session recording shortcut to Ctrl+Alt+V

lavjeetrai:
### What changed?
Changed the screen recording shortcut from `Ctrl+Shift+R` to `Shift+S`.

### Why?
`Ctrl+Shift+R` is the standard browser shortcut for a "Hard Refresh" (bypassing the cache). Using it as an application shortcut causes conflicts and triggers unexpected page reloads for users. Changing it to `Shift+S` resolves this issue while maintaining an accessible shortcut for screen recording.

### DIFF
--- a/packages/phoenix-event-display/src/event-display.ts
+++ b/packages/phoenix-event-display/src/event-display.ts
@@ -993,15 +993,15 @@ export class EventDisplay {
       document.removeEventListener('keydown', this.sessionRecordKeydownHandler);
     }
 
-    // Ctrl+Alt+V = toggle session recording
+    // Shift+S = toggle session recording
     this.sessionRecordKeydownHandler = (e: KeyboardEvent) => {
       const target = e.target as HTMLElement;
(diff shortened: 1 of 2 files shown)


rx18-eng:
Nice catch ! but Ctrl+Alt+V  may be not  the right replacement. Ctrl+Alt is the same as AltGr at the OS level, so  on non-US layouts where AltGr is used for everyday typing, so AltGr+V would fire this by accident mid-text. it's also Paste Special in a bunch of apps. Ctrl+Alt combos are generally avoided for exactly this reason.

could you switch it to Shift+S instead? S for session, it's a free key, matches the existing single-Shift shortcuts (rotate, clipping, theme...), and it dodges all three traps: the browser refresh, the AltGr clash, and Alt+Shift (which is the Windows language-switch combo).

lavjeetrai:
Thanks @rx18-eng I will update the shortcut key

rx18-eng (approved):

PR #946, opened 2026-06-27
 fix: restrict masterclass track selection to track-type collections 

### LINKED ISSUE #923: Masterclass kinematics has a "select tracks" pulldown which allows you to select non-track selections
I just spotted that the Kinematics MC window has a pulldown that starts with CSCs, which are not tracks. This could be a bit confusing for students - it would be better to restrict this to the track type (see e.g. the event data collections window for how this is retrieved).

Image

rx18-eng:
Closes #923.

## Summary

 the masterclass **"Select tracks"** dropdown was listing CSCs (muon chamber hits) and other non-track collections instead of just tracks, which is confusing for students who are supposed to be selecting tracks.

The bug was in `loadCollections()` in the masterclass overlay. `getCollections()` returns collections grouped by type (`Tracks`, `Hits`, `CaloClusters`, etc.), and the old code simply flattened every group into the dropdown, so hit collections like CSCs leaked in.

This PR filters the list to the `Tracks` group only, matching how the kinematics panel already resolves collections (`config.collectionType ?? 'Tracks'`).

To keep this extensible, I also added an optional `collectionTypes` field to `MasterclassConfig`, so experiments can explicitly opt other collection types back in if needed (for example, calorimeter clusters for photon-focused exercises where there are no tracks).

## Testing

- Added 4 regression tests covering the collection filtering logic.
- Verified manually on `/atlas-masterclass`:
  - the dropdown now lists only track collections
  - CSC and other non-track collections no longer appear

### DIFF
--- a/packages/phoenix-event-display/src/helpers/invariant-mass.ts
+++ b/packages/phoenix-event-display/src/helpers/invariant-mass.ts
@@ -57,6 +57,12 @@ export interface MasterclassConfig {
    * Returns a short label like "e", "4e", "2e2m".
    */
   classifyEvent: (tagCounts: Record<string, number>) => string;
+  /**
+   * Collection types shown in the "Select tracks" dropdown, e.g. ['Tracks'].
+   * Restricts the list to taggable types so detector-level collections (hits
+   * such as CSCs, calo cells) do not appear. Defaults to ['Tracks'] when unset.
+   */
+  collectionTypes?: string[];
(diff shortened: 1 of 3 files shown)


rx18-eng:
@EdwardMoyse  lmk your thoughts on this!

EdwardMoyse:
Looks good - thanks!

PR #951, opened 2026-07-02
Implement track cuts by distance of closest approach (DCA) and angle

raman118:
In certain experiments (such as LHCb), track collections are loaded containing only raw coordinate positions (pos) without track helix parameters (dparams). Because properties like $d_0$, $z_0$, $\phi$, and $\eta$ were not explicitly provided, these track cuts were completely filtered out from the user interface, preventing users from cutting away background tracks or parallel noise.

Solution
This PR introduces dynamic calculation of all missing track properties directly from their coordinates (pos), enabling the existing track cuts globally for all loaded track formats. It also introduces two new dedicated cuts specifically for filtering out parallel and displaced tracks:

DCA (Distance of Closest Approach) to the beamline.
Angle of the track relative to the beamline.
Mathematical Calculations
For any track, the tangent vector $\vec{v} = (v_x, v_y, v_z)$ is calculated using the first segment of the track (positions[1] - positions[0]):

Azimuthal Angle ($\phi$): Computed using Math.atan2(v_y, v_x).
Pseudorapidity ($\eta$): Computed from the polar angle $\theta$.
Transverse Impact Parameter ($d_0$): Calculated as the 2D distance of closest approach to the z-axis (beamline) in the XY plane: $$d_0 = \frac{x_0 v_y - y_0 v_x}{\sqrt{v_x^2 + v_y^2}}$$
Longitudinal Impact Parameter ($z_0$): Calculated at the point of closest approach in the XY plane: $$z_0 = z_0 - v_z \frac{x_0 v_x + y_0 v_y}{v_x^2 + v_y^2}$$
DCA: Computed as $|d_0|$.
Angle: Computed as the polar angle relative to the z-axis in degrees: $$\theta = \arccos\left(\frac{v_z}{|\vec{v}|}\right) \times \frac{180}{\pi}$$
Proposed Changes
packages/phoenix-event-display
src/loaders/objects/phoenix-objects.ts
Extracted track parameters logic into a reusable static method calculateTrackParams.
Integrated calculateTrackParams inside both getTrack and getTracks to automatically compute and assign dca, angle, d0, z0, phi, and eta from coordinate arrays if they are missing.
src/loaders/object-type-registry.ts
Registered the default dca (range 0 to 5000 mm) and angle (range 0 to 180 degrees) cuts in the track registry.
src/helpers/pretty-symbols.ts
Configured the mappings for dca $\rightarrow$ 'DCA' and angle $\rightarrow$ 'Angle' in the pretty symbols registry.
Verification & Testing
Unit Tests: Added a new test suite in phoenix-objects.test.ts verifying:
Exact property assignment of d0, z0, phi, eta, dca, and angle from track points.
Math correctness of signed transverse impact parameter ($d_0$), DCA, and polar angle.
Linting & Code Style: Clean ESLint check and auto-formatting applied. All 276 unit tests pass successfully.

### DIFF
--- a/packages/phoenix-event-display/src/helpers/pretty-symbols.ts
+++ b/packages/phoenix-event-display/src/helpers/pretty-symbols.ts
@@ -15,6 +15,8 @@ export class PrettySymbols {
     Energy: ['energy'],
     ET: ['et'],
     '|p|': ['momentum', 'mom'],
+    DCA: ['dca'],
+    Angle: ['angle'],
   };
 
   /**
--- a/packages/phoenix-event-display/src/loaders/object-type-registry.ts
+++ b/packages/phoenix-event-display/src/loaders/object-type-registry.ts
@@ -51,6 +51,8 @@ export function getDefaultObjectTypeConfigs(): ObjectTypeConfig[] {
         new Cut('pT', 0, 50000, 0.1),
         new Cut('z0', -30, 30, 0.1),
         new Cut('d0', -30, 30, 0.1),
+        new Cut('dca', 0, 5000, 0.1),
+        new Cut('angle', 0, 180, 0.1),
       ],
     },
     {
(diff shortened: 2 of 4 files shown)


EdwardMoyse:
This looks great - thanks!

PR #944, opened 2026-06-27
feat(phoenix-ng): add NotificationToastComponent — WA1 Step 3 UI layer for NotificationService

deveshbervar:
Closes WA1 Step 3.

Adds `NotificationToastComponent` — the UI layer for the`NotificationService` introduced in #895.

## What this adds
- `NotificationToastComponent` subscribes to `NotificationService` and displays toasts via `MatSnackBar`
- Severity levels map to CSS panel classes for styling
- Error notifications do not auto-dismiss (duration 0)
- Component added to `app.component.html` for global coverage
- 6 tests covering all 4 severity levels and unsubscribe cleanup

## WA1 arc
- Step 1 — #879 (merged) — JSON parse error handling
- Step 2 — #895 (merged) — NotificationService
- Step 3 — this PR — UI toast component

PR #955, opened 2026-07-07
feat(io-options-dialog): wire NotificationService to surface IO errors as toasts

deveshbervar:
Follow-up to #944 as suggested by @EdwardMoyse 

Wires `NotificationService` into `IOOptionsDialogComponent` so all existing IO errors now show both in the persistent Info Panel. 
AND as a transient toast visible immediately to the user.

## Error sites updated (6 total)
- JSON event file parse error
- EDM4HEP JSON parse error  
- PHYSLITE file load error
- Zip file read error
- FileReader onerror
- Invalid file extension error

## No behavior change
Existing `getInfoLogger().add()` calls are untouched — 
`notificationService.error()` is added alongside each one.


EdwardMoyse:
This is great! Thanks so much.

PR #959, opened 2026-07-12
Make cuts panel compact: inline checkbox+input per side, add scrollability

### LINKED ISSUE #953: More compact cut representation
We have the problem now that if you open the cuts on e.g. Tracks in ATLAS the cuts stretch off the bottom of the screen. We should probably make them scrollable, but as well I think we could make them a lot more compact (for example, having the name and ticks all in line with the values)

Image

Gitjay11:
## Summary

Fixes #953

The cuts panel for collections with many cuts (e.g. ATLAS Tracks with φ, η, χ², DOF, pT, z0, d0, DCA) was overflowing the bottom of the screen. This PR addresses both points raised in the issue:

1. **More compact layout** :  each cut now renders on a single row instead of two
2. **Scrollable panel** :  the cuts panel is now capped at 60vh and scrolls if needed

## Testing
- Verified all `ConfigRangeSlider` property bindings (`enableMin`, `enableMax`, `value`, `highValue`, `setEnableMin`, `setEnableMax`, `step`) match the type definition in `config-types.ts`
- No logic changes — only layout restructuring

### DIFF
--- a/packages/phoenix-ng/projects/phoenix-ui-components/lib/components/phoenix-menu/phoenix-menu-item/phoenix-menu-item.component.html
+++ b/packages/phoenix-ng/projects/phoenix-ui-components/lib/components/phoenix-menu/phoenix-menu-item/phoenix-menu-item.component.html
@@ -111,39 +111,41 @@
               />
 
               <div *ngSwitchCase="'rangeSlider'" class="range-slider">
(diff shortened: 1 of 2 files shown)


EdwardMoyse:
This is great! Thank you!

PR #964, opened 2026-07-15
fix(ui): revert cdkDrag on phoenix menu

### LINKED ISSUE #961: Phoenix menu dragged when selecting options.
After #954, when you click on something in the phoenix menu ("Colour Options" for example) you are now dragging the menu and cannot let go of it. This makes it impossible/very hard to use the UI.

@lavjeetrai could you take a look? If there is not a quick fix, i think we need to revert adding `cdkDrag` to the Phoenix menu.

Image

lavjeetrai:
fix #961 ,
For now, I have reverted the `cdkDrag` changes.


EdwardMoyse:
Thanks!

PR #973, opened 2026-08-07
refactor(effects-manager): extract GLSL shaders to TypeScript modules (#886)

deveshbervar:
## Summary

Phase 2a of the refactoring proposed in #886.

Extracts the two inline GLSL shader strings from`effects-manager.ts` into dedicated TypeScript modules.

## What changed

**New files:**
- `shaders/hover-vertex.ts` — vertex shader as TS default export
- `shaders/hover-fragment.ts` — fragment shader as TS default export

**Modified:**
- `effects-manager.ts` — imports shaders from modules instead of defining them as inline class properties

## Why TypeScript modules instead of .glsl files

Previous attempt (#933) used `.glsl` files with `raw-loader` but hit esbuild/Angular build pipeline issues in the Docker
smoke test. TypeScript modules work natively with all three build tools (webpack, esbuild, Jest) with zero configuration
changes.

## Scope
- 2 new files, 1 modified file
- Zero behavior change — shaders are identical
- All 25 effects-manager tests pass
- Part of #886 — Phase 2, PR 2a

### DIFF
--- a/packages/phoenix-event-display/src/managers/three-manager/shaders/hover-vertex.ts
+++ b/packages/phoenix-event-display/src/managers/three-manager/shaders/hover-vertex.ts
@@ -1,0 +1,8 @@
+/** Vertex shader for hover outline rendering. */
+const hoverVertexShader = `
+void main() {
+  gl_Position = projectionMatrix * modelViewMatrix * vec4(position, 1.0);
+}
+`;
+
+export default hoverVertexShader;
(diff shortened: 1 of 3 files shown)


packages/phoenix-event-display/src/managers/three-manager/shaders/hover-vertex.ts
```diff
@@ -0,0 +1,13 @@
+/** Vertex shader for hover outline rendering. */
+const hoverVertexShader = `
+uniform float opacity;
```
EdwardMoyse: Why are these lines added? These "uniform float" lines are not needed are they?

EdwardMoyse:
One minor comment - looks good otherwise. Thanks!

deveshbervar:
Good catch @EdwardMoyse — those uniform declarations in the vertex shader were redundant. The vertex shader only computes gl_Position and doesn't use those uniforms. Removed.

PR #977, opened 2026-08-20
feat(event-display): Allow vertices to be linked to different track t…

valponce:
Converts `linkedTracksCollection`  to a list of strings to allow tracks from different collections to be linked to the same vertex. Track IDs are now grouped in sublists by collection type inside `linkedTracks`.

### DIFF
--- a/packages/phoenix-event-display/src/lib/types/event-data.ts
+++ b/packages/phoenix-event-display/src/lib/types/event-data.ts
@@ -194,9 +194,9 @@ export interface VertexParams {
   /** Color for rendering. */
   color?: string | number;
   /** Indices of the tracks associated with this vertex, used to color tracks by vertex. */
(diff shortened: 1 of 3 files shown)


sponce:
Do we need this to be backward compatible ? If no, then I would at least rename `linkedTrackCollection` to `linkedTrackCollections` with an extra 's'. If yes, then we probably need new entry points (e.g. `linkedTracksPerCollection` and `linkedTrackCollections`) and support both old ones and new ones.
@EdwardMoyse what do you think ?

EdwardMoyse:
I'm a bit worried this will break deployed ATLAS configs e.g. for physics briefings. Will do some tests.

EdwardMoyse:
Sorry for the delay - I tested this on older configs and it seems to work fine.

Issue #985, opened 2026-09-04
anglesAreSane rejects valid tracks  and its test hides the bug


Shivansh1205:
The doc comment says 0 < theta < 2PI, but the code checks Math.PI > theta. The single test calls it as anglesAreSane(phi, theta) with phi=1.0, theta=3.0  arguments swapped relative to the signature (theta, phi). It passes only because 1.0 and 3.0 both happen to satisfy the swapped predicate. Called correctly as (3.0, 1.0) it returns false.

This is live in the JiveXML loader at [jivexml-loader.ts:326], where real tracks with theta between π and 2π get discarded as 'Improper angles'. Either the doc or the bound is wrong worth settling which, since it silently drops data.

PR #986, opened 2026-09-04
fix(event-display): correct anglesAreSane docs and test argument order

Shivansh1205:
## Problem

The doc comment for `CoordinateHelper.anglesAreSane` states that the accepted range is:

`-PI < phi < PI` and `0 < theta < 2PI`

However, the implementation checks:

```ts
const tmp2 = 0 < theta && Math.PI > theta;
```

which enforces `0 < theta < PI`.

The documentation and implementation therefore disagree about the valid range of `theta` by a factor of two.

## Which one is right?

The implementation is correct, and the documentation is wrong.

* `theta` is the polar angle and is conventionally defined in the range `[0, PI]`.
* `thetaToEta` computes `-log(tan(theta / 2))`, which is only real-valued for `theta` in `(0, PI)`.
* In `jivexml-loader.ts`, `theta` is calculated using `Math.atan(1 / cotTheta[i])`, producing values in `(-PI/2, PI/2)`.
* Negative values are then normalised using:

```ts
if (theta < 0) theta += Math.PI;
```

resulting in `theta` values within `(0, PI)`.

Therefore, `anglesAreSane` correctly enforces `0 < theta < PI`. Widening the check to `2PI` would allow invalid theta values through, which could subsequently cause pseudorapidity calculations to produce `NaN`.

## Why the test did not catch it

The existing test calls `anglesAreSane` with its arguments swapped relative to the function signature:

```ts
anglesAreSane(theta, phi)
```

The test currently does:

```ts
const phi = 1.0;
const theta = 3.0;

expect(CoordinateHelper.anglesAreSane(phi, theta)).toBe(true);
```

With `phi = 1.0` and `theta = 3.0`, both argument orderings happen to return `true`. As a result, the swap is invisible, and neither parameter's boundary conditions are actually covered.

The test would still pass even if either parameter's bounds were implemented incorrectly.

## Changes

* Correct the `anglesAreSane` doc comment to document the enforced range:

  * `-PI < phi < PI`
  * `0 < theta < PI`
* Add documentation explaining why `theta` is restricted to `(0, PI)`.
* Replace the single swapped-argument assertion with tests that:

  * Pass arguments in the declared order.
  * Exercise `phi` and `theta` independently.
  * Verify the exclusive boundaries at `0` and `PI`.
  * Verify the exclusive boundaries at `-PI` and `PI` for `phi`.

## Behaviour

No runtime behaviour is changed.

The `anglesAreSane` implementation is left untouched. This is strictly a documentation and test-correctness fix, so track classification remains unchanged.

### DIFF
--- a/packages/phoenix-event-display/src/helpers/coordinate-helper.ts
+++ b/packages/phoenix-event-display/src/helpers/coordinate-helper.ts
@@ -5,10 +5,13 @@ import { Vector3, Quaternion } from 'three';
  */
 export class CoordinateHelper {
   /**
-   * Checks if angles are within range: -PI < phi < PI and 0 < theta < 2PI
-   * @param theta equatorial angle
-   * @param phi azimuthal angle
-   * @returns
+   * Checks if angles are within range: 0 < theta < PI and -PI < phi < PI.
+   * Theta is the polar angle, which is only defined on [0, PI] - values
+   * outside it have no real pseudorapidity, since thetaToEta computes
+   * -log(tan(theta / 2)).
+   * @param theta polar angle in radians, expected in (0, PI)
+   * @param phi azimuthal angle in radians, expected in (-PI, PI)
+   * @returns true if both angles are within their expected ranges
    */
   public static anglesAreSane(theta: number, phi: number): boolean {
     const tmp1 = -Math.PI < phi && Math.PI > phi;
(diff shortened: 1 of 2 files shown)


Shivansh1205:
@EdwardMoyse kindly review its for the issue #985

EdwardMoyse:
Thanks very much for this!

Issue #987, opened 2026-09-04
Cartesian grid crashes on non-default scales — scene-manager.ts:780


Shivansh1205:
for (let i = 0; i <= 62; i += 1) this.cartesianGrid.children[i].visible = false; hardcodes 63 children. The count actually comes from float-accumulating loops (z += 0.1 * scale) in createCartesianGrid, so it is scale-dependent. I verified numerically:

scale=3000 → 63 children   (works)
scale=7    → 60 children   (children[60] is undefined → TypeError)
scale=0.1  → 60 children
Default scale is 3000, so it works today, but any non-default scale throws on .visible. The magic numbers [10, 31, 52] just below share the assumption.

PR #988, opened 2026-09-04
fix(event-display): make cartesian grid work at non-default scales

Shivansh1205:
## Problem

`setCartesianGrid` iterated over `cartesianGrid.children` using:

```ts
for (let i = 0; i <= 62; i += 1)
```

This hardcoded the expectation of 63 planes and derived visibility offsets from the magic indices `[10, 31, 52]`.

That count is not guaranteed to be fixed. The planes were generated using floating-point accumulation (`z += 0.1 * scale`), so rounding could determine whether the final plane was emitted. For example:

* Scale `3000` → 21 planes per axis → 63 total.
* Scale `7` or `0.1` → 20 planes per axis → 60 total.

At those scales, `children[60]` and subsequent entries are undefined, causing `.visible` access to throw a `TypeError`.

The default scale is `3000`, making this a latent issue that only becomes apparent when using a non-default scale.

The visibility loops had a related issue: `j` was incremented by `sparsity` without clamping. A sufficiently large distance relative to the scale could cause the index to leave its axis' own block, enter a neighbouring block, or exceed the array bounds.

## Changes

* Drive plane-generation loops using an integer step index instead of accumulating floating-point coordinates.
* Ensure each axis always generates exactly `2 * CARTESIAN_GRID_STEPS + 1` planes at every scale.
* Derive the per-axis block centres from `CARTESIAN_GRID_STEPS` instead of hardcoding `[10, 31, 52]`.
* Clamp each axis' visibility range to its own block so indices cannot enter neighbouring blocks or exceed the array bounds.
* Guard against zero or negative `sparsity`, which would otherwise prevent the visibility loop from advancing.
* Apply the same integer-step approach to the cartesian label loops, avoiding floating-point accumulation that could cause tick labels to be skipped at certain scales.

## Verification

Verified that visibility is unchanged at the default scale of `3000` across all 96 combinations of distance, sparsity, and per-plane toggles.

This fixes the non-default-scale failures and index-boundary issues without changing existing visual behaviour at the default scale.

Fixes (https://github.com/HSF/phoenix/issues/987)](https://github.com/HSF/phoenix/issues/987)

### DIFF
--- a/packages/phoenix-event-display/src/managers/three-manager/scene-manager.ts
+++ b/packages/phoenix-event-display/src/managers/three-manager/scene-manager.ts
@@ -57,6 +57,15 @@ export class SceneManager {
   private etaPhiGrid: Object3D;
   /** Cartesian grid */
   private cartesianGrid: Object3D;
+  /**
+   * Number of grid steps from the centre of the cartesian grid to each edge.
+   * The grid spans [-scale, scale] in steps of `0.1 * scale`, so each axis has
+   * `2 * CARTESIAN_GRID_STEPS + 1` planes (the centre plane plus both halves).
+   */
+  private static readonly CARTESIAN_GRID_STEPS = 10;
+  /** Number of planes built per axis, derived from CARTESIAN_GRID_STEPS. */
+  private static readonly CARTESIAN_PLANES_PER_AXIS =
+    2 * SceneManager.CARTESIAN_GRID_STEPS + 1;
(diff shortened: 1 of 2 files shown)


Shivansh1205:
@EdwardMoyse kindly review

EdwardMoyse (approved):
This looks like a comprehensive fix to a real problem. Thank you!

PR #992, opened 2026-09-04
fix: revoke object URLs after use

### LINKED ISSUE #990: Object URLs never revoked for ROOT files — io-options-dialog.component.ts:223 and :238
URL.createObjectURL(files[0]) with no matching revokeObjectURL, pinning the entire file in memory for the page's lifetime. ROOT/PHYSLITE files are routinely hundreds of MB. Same defect in file.ts:15 (saveFile). This is clearly an oversight rather than a convention — three other call sites (session-pill, kinematics-panel, masterclass-panel, three-manager) all revoke correctly.

Shivansh1205:
## Problem

`URL.createObjectURL` keeps the underlying `Blob` or `File` alive until the object URL is revoked. If the URL is never revoked, the entire file remains pinned in memory for the lifetime of the page.

`handleROOTInput` and `handleRootJSONInput` both created object URLs for selected files but never revoked them. ROOT and PHYSLITE files are routinely hundreds of MB, so loading multiple geometries could leak a significant amount of memory until the tab was reloaded.

`saveFile` had the same issue with the `Blob` used for downloads.

## Changes

* Revoke the object URL in a `finally` block in both dialog handlers.

  * This ensures the URL is released whether geometry loading succeeds or throws.
  * Both loaders are asynchronous and finish reading the URL before resolving, so revoking it in `finally` cannot happen prematurely.
* Update `saveFile` to revoke its object URL on the next tick rather than synchronously.

  * The anchor click starts the browser download asynchronously.
  * Revoking immediately after the click can cancel the download.
  * Deferring the revocation gives the browser time to retain its own reference before the URL is released.

## Verification

Added tests covering `saveFile` to verify that:

* The object URL created by `saveFile` is eventually revoked.
* The URL is not revoked synchronously, allowing the download to start successfully.

This addresses an object URL lifecycle oversight. Other call sites, including `session-pill`, `kinematics-panel`, `masterclass-panel`, and `three-manager`, already revoke their object URLs correctly.

Fixes #990

### DIFF
--- a/packages/phoenix-event-display/src/helpers/file.ts
+++ b/packages/phoenix-event-display/src/helpers/file.ts
@@ -12,10 +12,15 @@ export const saveFile = (
   const blob = new Blob([data], { type: contentType });
   const tempAnchor = document.createElement('a');
   tempAnchor.style.display = 'none';
-  tempAnchor.href = URL.createObjectURL(blob);
+  const objectUrl = URL.createObjectURL(blob);
+  tempAnchor.href = objectUrl;
   tempAnchor.download = fileName;
   tempAnchor.click();
   tempAnchor.remove();
+  // The download is started asynchronously by the click above, so revoking
+  // synchronously here can cancel it. Release the URL on the next tick, once
+  // the browser has taken its own reference to the blob.
+  setTimeout(() => URL.revokeObjectURL(objectUrl));
 };
 
 /**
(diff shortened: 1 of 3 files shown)


Shivansh1205:
@EdwardMoyse kindly review

EdwardMoyse:
Looks good - thanks for this!

PR #993, opened 2026-09-04
fix(ui-components): tear down grid-shift subscriptions on destroy

### LINKED ISSUE #989: Grid-shift subscription leak / TypeError on race — cartesian-grid-config.component.ts:57-69
Two subscriptions unsubscribe only inside the stopShifting handler, which fires on right-click (index.ts:740). The component calls this.onClose() immediately after subscribing, and has no ngOnDestroy. Close the dialog without right-clicking and both subscriptions outlive it, leaking on every reopen. If stopShifting emits before stopShiftingSub is assigned, line 67 throws on null.

Shivansh1205:
## Problem

`shiftCartesianGridByPointer` subscribed to `originChanged` and `stopShifting`, then immediately called `onClose()` to dismiss the dialog. Both subscriptions were only unsubscribed from inside the `stopShifting` handler, which fires on right-click, and the component did not implement `ngOnDestroy`.

As a result, closing the dialog without right-clicking left both subscriptions active for the lifetime of the page. Reopening the dialog accumulated another pair of subscriptions each time, and the stale `originChanged` handler could continue translating the grid after the component had been destroyed.

The teardown also had a synchronous emission bug. `stopShifting` can emit synchronously during `subscribe`, causing its handler to run before `stopShiftingSub` has been assigned. Calling `this.stopShiftingSub.unsubscribe()` in that situation throws a `TypeError`. The component's existing test mocks use `of(true)`, which emits synchronously and exposes this issue.

## Changes

* Implement `OnDestroy` so all active shift subscriptions are cleaned up when the component is destroyed.
* Introduce a single `clearShiftSubscriptions` helper responsible for:

  * Unsubscribing from both subscriptions when present.
  * Nulling the subscription fields after teardown.
  * Being safe to call repeatedly.
  * Being safe to call before a subscription has been assigned.
  * Being safe to call when no shift is currently active.
* Clear any existing shift subscriptions before starting a new shift, preventing repeated shifts from accumulating subscriptions.
* Handle synchronous `stopShifting` emissions using a local flag rather than checking the subscription's `closed` state. A plain synchronous observable does not necessarily report the subscription as closed at the point the handler executes.
* Update the existing `should shift cartesian grid by a mouse click` test.

  * The previous test placed its expectations inside `subscribe` callbacks, so the assertions were never actually executed.
  * The test now verifies the observable behaviour directly.

## Verification

The updated teardown logic safely handles component destruction, repeated shifts, synchronous observable emissions, and normal stop events without leaking subscriptions.

Fixes #989

### DIFF
--- a/packages/phoenix-ng/projects/phoenix-ui-components/lib/components/ui-menu/view-options/cartesian-grid-config/cartesian-grid-config.component.test.ts
+++ b/packages/phoenix-ng/projects/phoenix-ui-components/lib/components/ui-menu/view-options/cartesian-grid-config/cartesian-grid-config.component.test.ts
@@ -40,10 +40,23 @@ describe('CartesianGridConfigComponent', () => {
     originChangedEmit: jest.fn().mockReturnThis(),
   };
 
+  // Several tests swap in Subjects so the subscriptions stay open; put the
+  // default completing observables back afterwards.
+  const restoreObservables = () => {
+    mockEventDisplay.originChanged = of(gridOrigin) as any;
+    mockEventDisplay.stopShifting = of(true) as any;
+  };
+
   const mockData = {
     gridVisible: true,
     scale: 3000,
   };
+
+  afterEach(() => {
+    // Restore the default completing observables, even if a test failed part
+    // way through, so a swapped-in Subject cannot leak into the next test.
+    restoreObservables();
+  });
 
   beforeEach(async () => {
     await TestBed.configureTestingModule({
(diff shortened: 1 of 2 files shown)


EdwardMoyse:
Please run `yarn lint:fix`

packages/phoenix-ng/projects/phoenix-ui-components/lib/components/ui-menu/view-options/cartesian-grid-config/cartesian-grid-config.component.test.ts
```diff
@@ -6,6 +6,7 @@ import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material/dialog';
 import { MatCheckboxChange } from '@angular/material/checkbox';
 import { Vector3 } from 'three';
 import { of } from 'rxjs/internal/observable/of';
+import { Subject } from 'rxjs';
```
EdwardMoyse, line 9: I might be missing something here, but where is this actually used?

Shivansh1205: Fair question  at the commit that diff line came from [b8af2fa8], a Prettier fixup the import genuinely was unused. It gets used in the next commit, [dc1394c2], which is on the PR head.

The reason it's needed: the existing mocks use of(...), which emits and completes immediately, so the subscription auto-closes on its own. That masks exactly the leak this PR fixes  the test would pass whether or not the teardown works. Subject never completes by itself, so the subscriptions stay live and the teardown becomes observable.

PR #994, opened 2026-09-04
fix(event-display): handle malformed JSON in user-loaded files

### LINKED ISSUE #991: Unguarded JSON.parse on user-supplied files — state-manager.ts:77 and ui-manager/index.ts:588
Both parse user-chosen files inside a loadFile callback with no try/catch, so a malformed file throws an unhandled exception with no user feedback. Notably, session-codec.ts does this correctly with proper error messages — these two just weren't updated to match.

Shivansh1205:
## Problem

`Load state` and `Load labels` both call `JSON.parse` on the contents of a user-selected file inside a `loadFile` callback without any error handling.

If the selected file contains invalid JSON, `JSON.parse` throws a `SyntaxError` from inside the callback. Nothing catches the error, so the load silently fails and the user receives no indication of what went wrong.

`session-codec.ts` already guards its JSON parsing this way, so these two call sites were simply not updated to follow the same pattern.

## Changes

* Wrap the `JSON.parse` calls in both `Load state` and `Load labels` with error handling.
* Report malformed JSON via `console.error`, matching the error-handling behaviour used by the surrounding managers.
* Return immediately after a parse failure so invalid data is never applied.
* Leave valid state and label loading behaviour unchanged.

## Verification

Added a test covering malformed state files to verify that:

* The parsing failure is reported instead of escaping as an uncaught exception.
* `loadStateFromJSON` is not called with invalid/garbage data.

This brings the two file-loading paths in line with the existing error-handling approach in `session-codec.ts`.

Fixes #991

### DIFF
--- a/packages/phoenix-event-display/src/managers/state-manager.ts
+++ b/packages/phoenix-event-display/src/managers/state-manager.ts
@@ -74,7 +74,19 @@ export class StateManager {
           label: 'Load state',
           onClick: () => {
             loadFile((data) => {
-              this.loadStateFromJSON(JSON.parse(data));
+              // The file comes from the user, so it may not be valid JSON.
+              // Report the problem instead of throwing out of the callback.
+              let state: any;
+              try {
+                state = JSON.parse(data);
+              } catch (error) {
+                console.error(
+                  'Could not parse state file - invalid JSON.',
+                  error,
+                );
+                return;
+              }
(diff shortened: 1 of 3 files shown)


Shivansh1205:
@EdwardMoyse kindly review

EdwardMoyse:
Thanks!

PR #998, opened 2026-09-06
build(deps): resolve vulnerable transitive tar dependency

amiitt001:
## Summary

Update the transitive `tar` dependency resolution to remove vulnerable
`tar@6.x` versions from the repository dependency tree.

## Changes

- Added a root Yarn resolution for `tar` at `^7.5.19`.
- Regenerated `yarn.lock`.
- No application source code was changed.

## Verification

- `corepack yarn why tar` confirms the dependency tree resolves to the patched `tar` release.
- No `tar@6.x` instances remain in the dependency tree.
- `corepack yarn test:ci` passes across all projects.
- `corepack yarn lint` passes with no errors.
- `corepack yarn lerna run build` completes successfully.
- Only `package.json` and `yarn.lock` were modified.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,8 @@
     "typescript": "~5.8.2"
   },
   "resolutions": {
-    "three": "~0.178.0"
+    "three": "~0.178.0",
+    "tar": "^7.5.19"
   },
   "packageManager": "yarn@3.3.1",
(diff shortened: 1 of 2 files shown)


EdwardMoyse:
Thank you very much for this PR.

PR #1029, opened 2026-09-13
fix(loaders): handle missing tracker hits and fallback to track state

lavjeetrai:
### Description
Fixes an issue where missing `trackerHits` collections or out-of-bound hit indices cause unhandled `TypeError` exceptions during EDM4hep JSON event loading. Additionally, enables graceful fallback to `trackStates` when tracker hit coordinates cannot be resolved.

### Changes Made
- **Null Safety:** Guarded `trackerHits` and `trackStates` array access and sub-properties (`hit?.position`, `trackState?.referencePoint`).
- **Dynamic Fallback:** Decoupled `trackerHits` and `trackStates` branches so tracks with unresolvable hit references fall back to `trackStates` (`if (pos.length === 0 && ...)`).
- **Refactoring:** Cached indexed hit lookup to avoid duplicate collection scans and redundant array indexing.
- **Unit Tests:** Added full test suite in `packages/phoenix-event-display/src/tests/loaders/edm4hep-json-loader.test.ts`.

### Test Plan
- [x] Tested with valid `trackerHits` references.
- [x] Tested fallback when `collectionID` is missing from `rawEvent`.
- [x] Tested fallback when hit index is out of bounds.
- [x] Tested track loading with only `trackStates`.
- [x] Tested empty track edge cases without throwing errors.


lavjeetrai:
@EdwardMoyse could u take a look to this.

packages/phoenix-event-display/src/loaders/edm4hep-json-loader.ts
```diff
@@ -210,27 +210,33 @@ export class Edm4hepJsonLoader extends PhoenixLoader {
     trackCollection.forEach((rawTrack: edm4hep.Track) => {
       const pos: number[][] = []; // An array of positions is needed to render the tracks as bars
 
-      // @todo trackerhits might always exist
-      if ('trackerHits' in rawTrack && rawTrack.trackerHits.length > 0) {
+      if (rawTrack.trackerHits?.length > 0) {
```
EdwardMoyse: Isn't this less protection than before? Do we know that `trackerHits` are guaranteed to exist in `rawTrack`? What is the harm in checking?

lavjeetrai:
@EdwardMoyse please take a look again

PR #767, opened 2026-01-01
Handle FileReader errors in IO Options dialog to prevent silent failures

remo-lab:
## What this PR fixes
This PR adds missing error handling for `FileReader `failures in the` handleFileInput` method used by the IO Options dialog.
Previously, when `FileReader` encountered an error (for example due to large files, corrupted content, or access issues), the dialog would close silently with no feedback to the user. From the user’s perspective, the file simply appeared to be ignored.
This change ensures that such failures are properly handled and communicated.

## Why this change is needed
Loading event data and geometry files is a core user workflow in Phoenix. The current behavior makes it difficult for users to understand why a file did not load, and it complicates debugging when something goes wrong.

This issue affects multiple file-loading paths and can occur in realistic scenarios such as:
- very large JSON / OBJ files,
- partially downloaded or corrupted files,
- browser memory or read failures.
Failing silently in these cases leads to poor UX and unnecessary confusion.

## What was changed
- Added a `FileReader.onerror handler `in `handleFileInput`
- When a read error occurs:
    - a clear error message is shown to the user via the existing `InfoLogger`
    -  technical details are logged to the browser console for debugging
- No changes were made to successful file-loading behavior

This follows the existing error-handling pattern already used in the component and introduces no new dependencies.

### Before
- FileReader errors caused silent failures
- Dialog closed with no indication of what went wrong
- Users had no feedback in the UI

### After
- FileReader errors are explicitly handled
- Users see a clear error message in the Info panel
- Developers can inspect console logs for details

This improves robustness and usability without introducing breaking changes.

## Affected workflows
This fix benefits all file loaders that rely on `handleFileInput`, including:
- JSON event data
- EDM4HEP JSON
- JiveXML
- OBJ geometry
- Phoenix scene files (`.phnx`)
- Scene files

## Scope and risk
- Small, localized change
- Purely additive error handling
- No API or behavior changes for successful reads
- Low risk and easy to reason about

PR #1033, opened 2026-09-14
fix(ui-components): report invalid .phnx files instead of failing silently

### LINKED ISSUE #1032: Invalid .phnx files fail silently: unhandled promise rejection in handleSceneInput / handlePhoenixInput
Loading a malformed `.phnx` file does nothing at all — no geometry, no error, no message. The parse failure becomes an unhandled promise rejection that no code observes.

## The offending code

Both `.phnx` handlers call `parsePhoenixDisplay` from inside the `FileReader.onload` callback, with nothing handling a failure:

- [`io-options-dialog.component.ts#L196-L201`](https://github.com/HSF/phoenix/blob/main/packages/phoenix-ng/projects/phoenix-ui-components/lib/components/ui-menu/io-options/io-options-dialog/io-options-dialog.component.ts#L196-L201) — `handleSceneInput`
- [`io-options-dialog.component.ts#L212-L217`](https://github.com/HSF/phoenix/blob/main/packages/phoenix-ng/projects/phoenix-ui-components/lib/components/ui-menu/io-options/io-options-dialog/io-options-dialog.component.ts#L212-L217) — `handlePhoenixInput`

```ts
handlePhoenixInput(files: FileList) {
  const callback = (content: any) => {
    this.eventDisplay.parsePhoenixDisplay(content);   // not awaited, not caught
  };
  this.handleFileInput(files[0], 'phnx', callback);
}
```

`parsePhoenixDisplay` parses the raw text directly:

https://github.com/HSF/phoenix/blob/main/packages/phoenix-event-display/src/event-display.ts#L555-L556

```ts
public async parsePhoenixDisplay(input: any): Promise<void> {
  const phoenixScene = JSON.parse(input);
```

## Why it is wrong

Two things combine:

1. `parsePhoenixDisplay` is `async`, so a `JSON.parse` failure does **not** throw synchronously — it returns a rejected promise.
2. The call site neither awaits it nor attaches a `.catch`, and it runs inside `reader.onload`.

The result is an unhandled rejection that no code observes. `reader.onerror` does not help: it fires when the *read* fails, not when the onload callback rejects. [`handleFileInput`](https://github.com/HSF/phoenix/blob/main/packages/phoenix-ng/projects/phoenix-ui-components/lib/components/ui-menu/io-options/io-options-dialog/io-options-dialog.component.ts#L385-L393) wires up `onerror` for read failures only, which is what #767 addressed — this is the other failure mode.

This is also inconsistent within the same file: the sibling JSON handlers already guard their parses and report the problem via the info logger and a notification.

```ts
// handleJSONEventDataInput — already does the right thing
try {
  const json = typeof content === 'string' ? JSON.parse(content) : content;
  this.eventDisplay.parsePhoenixEvents(json);
} catch (error) {
  this.notificationService.error('Could not parse JSON event file. …');
}
```

The `.phnx` handlers were simply never given the same treatment.

## What the user sees

Pick a `.phnx` file via **Load scene** or **Load Phoenix file**, and the dialog closes with no geometry loaded and no explanation. Realistic ways to land here: a truncated or interrupted download, a file saved from a different export, or an HTML error page saved with a `.phnx` extension. The only trace is an unhandled rejection in the devtools console, which a physicist using the app will not be looking at.

## How severe

It is not a crash in the browser, but it is a dead end with zero feedback — the user cannot tell whether the app is broken, the file is wrong, or the click was missed. In a Jest environment the same unhandled rejection is fatal: it aborts the worker process outright.

## Steps to reproduce

1. Save any non-JSON text (e.g. `not json`) as `broken.phnx`.
2. Open **IO Options** → **Load scene** (or **Load Phoenix file**) and pick it.
3. The dialog closes, nothing loads, and no error is shown anywhere in the UI.

## Expected behaviour

- A malformed `.phnx` file reports an error through the info logger and a notification, the same way an invalid `.json` event file already does.
- The failure never becomes an unhandled promise rejection.
- A valid `.phnx` file still loads exactly as before.

Worth noting for whoever reviews: the existing tests for these handlers mock `handleFileInput` wholesale, so the callback never runs and this path is currently untested — which is why the bug is invisible to CI.

I have a branch with the fix and tests, and will open a PR referencing this issue.

Shivansh1205:
Fixes #1032

## Problem

Loading a malformed `.phnx` file does nothing at all — no geometry, no error, no message.

Two things combine to swallow the failure:

1. `parsePhoenixDisplay` is `async` and calls `JSON.parse` on the raw text, so a bad file produces a **rejected promise**, not a synchronous throw.
2. `handleSceneInput` and `handlePhoenixInput` call it from inside the `FileReader.onload` callback without awaiting it or attaching a `.catch`.

Nothing observes the rejection. `reader.onerror` does not help — it fires when the *read* fails, not when the onload callback rejects. That read-failure path is what #767 added; this is the other failure mode.

It is also inconsistent within this very file: the sibling JSON handlers (`handleJSONEventDataInput`, `handleEDM4HEPJSONEventDataInput`) already catch malformed input and report it through the info logger plus a notification. The `.phnx` handlers were never given the same treatment.

## Fix

Route both handlers through a `parsePhoenixFile` helper that reports the failure the same way the JSON handlers already do.

The helper deliberately handles **both** arrival paths:

```ts
try {
  Promise.resolve(this.eventDisplay.parsePhoenixDisplay(content)).catch(onError);
} catch (error) {
  onError(error);
}
```

- `.catch` on the promise covers the actual bug — the async rejection.
- The surrounding `try` covers anything thrown before the first await, so this stays correct if `parsePhoenixDisplay` ever becomes synchronous.

`Promise.resolve(...)` rather than `await` keeps the callback non-async, so `handleFileInput`'s signature is untouched.

I kept the fix at the call site rather than adding a `try/catch` inside `parsePhoenixDisplay`, because that method has no access to the notification service and its other callers may want the rejection to propagate. This matches where the existing JSON guards live.

## Tests

Added three cases to the existing suite:

- invalid `.phnx` via **Load scene** → reports an error, does not throw
- invalid `.phnx` via **Load Phoenix file** → reports an error, does not throw
- valid `.phnx` → still loads, and reports nothing

The existing tests for these handlers mock `handleFileInput` wholesale, so the callback never runs and this path was entirely untested — which is why the bug was invisible to CI. The new tests capture the callback that the handler passes to `handleFileInput` and invoke it directly, exercising the real parse path without needing a live `FileReader`.

The mock is `async (content) => JSON.parse(content)`, reproducing the real method's shape: the parse failure arrives as a rejection, not a sync throw. An earlier draft of these tests used a synchronously-throwing mock and passed against a `try/catch`-only fix — the async mock is what actually pins the bug.

**Before this change** the two invalid-file tests do not merely fail an assertion: the unhandled rejection aborts the Jest worker outright.

```
SyntaxError: Unexpected token 'o', "not json" is not valid JSON
    at JSON.parse (<anonymous>)
    ...
Node.js v24.14.0
```

(no test summary is printed — the process dies)

After the fix, all 16 tests in the suite pass.

## Verification

- Full `phoenix-ng` suite: **62 suites / 224 tests passed**, 0 failures.
- `eslint` clean on the `io-options` directory.
- No pre-existing failures found, so none are carried here.


EdwardMoyse:
Thank you very much for this, and the other PRs!

